PR #61, opened 2019-12-18
Add CVV and AVS types, add CyberSource translations.

jakechiang-bolt:


nirajjayantbolt:
To confirm: Hail needs to deal with processor code directly so the Processor->Sleet won't be used by us but I think its not a bad feature to have gnerally speaking

PR #68, opened 2020-04-08
[Stripe] - clean up all of this code - use library

nirajjayantbolt:


nirajjayantbolt:
I dont remember why we didnt do this initially but we should be using the Golang library for all of this stuff

PR #77, opened 2020-04-23
[Adyen] - Adding AVS results

nirajjayantbolt:
adyen-api-go didnt support AVS so I forked, then fixed it so now lets use it!

### DIFF
--- a/gateways/adyen/translation.go
+++ b/gateways/adyen/translation.go
@@ -1,0 +1,49 @@
+package adyen
+
+import "github.com/BoltApp/sleet"
+
+// translateCvv converts a Adyen CVV response code to its equivalent Sleet standard code.
+func translateCvv(rawCvv string) sleet.CVVResponse {
+	// Codes taken from: https://docs.adyen.com/development-resources/test-cards/cvc-cvv-result-testing
+	switch rawCvv {
+	case "0":
+		return sleet.CVVResponseNoResponse
+	case "5":
+		return sleet.CVVResponseUnsupported
+	case "1":
+		return sleet.CVVResponseMatch
+	case "2":
+		return sleet.CVVResponseNoMatch
+	case "3", "6":
+		return sleet.CVVResponseNotProcessed
+	case "4":
+		return sleet.CVVResponseRequiredButMissing
+	default:
+		return sleet.CVVResponseUnknown
+	}
+}
+
+// translateAvs converts a Adyen AVS response code to its equivalent Sleet standard code.
+func translateAvs(rawAvs string) sleet.AVSResponse {
+	// Codes taken from: https://docs.adyen.com/risk-management/avs-checks
+	switch rawAvs {
+	case "A":
+		return sleet.AVSResponseZipNoMatchAddressMatch
+	case "N":
+		return sleet.AVSResponseNoMatch
+	case "R", "S", "U":
+		return sleet.AVSResponseUnsupported
+	case "W", "Z", "T":
+		return sleet.AVSResponseUnsupported
+	case "D", "F", "M", "X", "Y":
+		return sleet.AVSResponseMatch
+	case "B":
+		return sleet.AVSResponseZipUnverifiedAddressMatch
+	case "P":
+		return sleet.AVSResponseZipMatchAddressUnverified
+	case "C", "G", "I":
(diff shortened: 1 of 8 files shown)


nirajjayantbolt:
Actually let me add a new test while I've beefed it up

gateways/adyen/translation.go
```diff
@@ -0,0 +1,49 @@
...
+		return sleet.AVSResponseUnsupported
+	case "D", "F", "M", "X", "Y":
+		return sleet.AVSResponseMatch
+	case "B":
+		return sleet.AVSResponseZipUnverifiedAddressMatch
+	case "P":
+		return sleet.AVSResponseZipMatchAddressUnverified
+	case "C", "G", "I":
+		return sleet.AVSResponseSkipped
+	default: // Includes "2"
```
jakechiang-bolt, line 46: I think this comment is a holdover from CyberSource.

gateways/adyen/translation.go
```diff
@@ -0,0 +1,49 @@
...
+	// Codes taken from: https://docs.adyen.com/risk-management/avs-checks
+	switch rawAvs {
+	case "A":
+		return sleet.AVSResponseZipNoMatchAddressMatch
+	case "N":
+		return sleet.AVSResponseNoMatch
+	case "R", "S", "U":
+		return sleet.AVSResponseUnsupported
+	case "W", "Z", "T":
+		return sleet.AVSResponseUnsupported
```
jakechiang-bolt, line 37: We should add AVSResponseZipMatchAddressNoMatch for this case.

gateways/adyen/translation.go
```diff
@@ -0,0 +1,49 @@
...
+	default:
+		return sleet.CVVResponseUnknown
+	}
+}
+
+// translateAvs converts a Adyen AVS response code to its equivalent Sleet standard code.
+func translateAvs(rawAvs string) sleet.AVSResponse {
+	// Codes taken from: https://docs.adyen.com/risk-management/avs-checks
+	switch rawAvs {
+	case "A":
```
jakechiang-bolt, line 30: Is there a way to get Adyen's mapped response code instead of the raw code? If not we'll probably need to get the processor from the BIN so we can differentiate duplicate-meaning codes like "A".

nirajjayantbolt:
@jakechiang-bolt ah interesting point...yeah when I was doing this I kind of thought that this was dumb because hail doesn't consume any of the Sleet translated stuff

WDYT about just axing that and always returning a raw response? I think for our own posterity in hail we should always have raw creds which makes Sleet translations pretty useless

jakechiang-bolt:
@nirajjayantbolt So return only the raw response? As far as our use case for AVS/CVV is concerned I think that's fine, though I still like the idea of having a Sleet standard since it aligns with the PSP abstraction that Sleet's going for. I suppose it depends on whether we'd want to support a feature we don't actually use ourselves. What do we use raw results for in Hail aside from risk?

nirajjayantbolt:
@jakechiang-bolt yeah that makes sense you are right
I think it gets fed into our risk model and analysts look at it. We also have metrics related to auth retry that look at things like previous AVS result, etc but not sure if we could have a standard there.

nirajjayantbolt:
> Like I'm thinking that we ensure the Sleet AVS/CVV result types are a superset of all possible responses so all cases are handled, then all gateway translators can live in Sleet and we'll only need N+1 translations (gateways + Sleet->Hail) instead of 2N (gateways->Hail)

I like this a lot - we can even move on Cybs now before we have anyone live?

nirajjayantbolt:
@jakechiang-bolt although one problem is if we do translate it from Sleet, there is like literally no way of getting this back. We don't log wrapped client requests so we cant go to Athena. Not sure if that's a big deal but something to consider

nirajjayantbolt:
Also, then if we do the translation here then arent we kind of forcing Sleet to hold Bolt business logic?

jakechiang-bolt:
<Discussed offline> Keeping both translations for now.

gateways/adyen/translation.go
nirajjayantbolt: @jakechiang-bolt this is done now - I strongly typed what I've found from Adyen's codes and playing with postman

PR #78, opened 2020-04-27
Braintree + Gofmt of project

nirajjayantbolt:
Haha I noticed Braintree was just empty so filling it out and adding robust tests


currency_list.go
```diff
@@ -0,0 +1,170 @@
+package sleet
+
+var CURRENCIES = map[Code]Currency{
```
darshit-bolt, line 3: Is this list going to be used across all gateways? We need to check if all gateways follow the same ISO standard. For example Adyen differs from ISO on decimal precision for some currencies: https://docs.adyen.com/development-resources/currency-codes#note
Also it represents Serbian dinar as CSD and ISO represents it as RSD.

nirajjayantbolt: Actually that is a good question and something I realized I should probably adjust in the Adyen code

They have their own representation stored in adyen-api-go/amount.go which I should be using. So the question is more - how do we as Bolt/sleet want to represent it? So this I just copied from hail

nirajjayantbolt: Also how did you find that so quickly haha

darshit-bolt: Haha, I was previously reading some Adyen docs and came across the note which says they differ from ISO. Then I just diff a diff between our list and theirs and saw a few differences, lol.

Ya, I think it makes sense to make changes in the Adyen code, but we'll have to make sure we do corresponding changes in other gateways if they differ. Braintree seems fine though.

nirajjayantbolt: Yeah definitely noted - hmm then do you have an opinion on the stance we should take with Sleet?

darshit-bolt: Ya I think this PR does the right thing, we need a generic standard. If a gateway has some differences from ISO standard, we can write a gateway specific currency/amount adjustment logic.

integration-tests/braintree_test.go
```diff
@@ -31,30 +54,154 @@ func TestBraintree(t *testing.T) {
...
+	captureRequest := &sleet.CaptureRequest{
+		Amount:               &authRequest.Amount,
+		TransactionReference: auth.TransactionReference,
+	}
+	capture, err := client.Capture(captureRequest)
+	if err != nil {
+		t.Error("Capture request should not have failed")
+	}
+
+	if capture.Success == false {
```
jakechiang-bolt, line 84: Nit: !capture.Success (same for the 6 below)

gateways/braintree/request_builders.go
```diff
@@ -3,27 +3,47 @@ package braintree
...
+			Locality:          common.SafeStr(billingAddress.Locality),
+			Region:            common.SafeStr(billingAddress.RegionCode),
+			PostalCode:        common.SafeStr(billingAddress.PostalCode),
+			CountryCodeAlpha2: common.SafeStr(billingAddress.CountryCode),
+		}
+	}
+	return request, nil
+}
+
+func convertToBraintreeDecimal(amount int64, currencyCode string) (*braintree_go.Decimal, error) {
```
darshit-bolt, line 42: This seems gateway agnostic, so probably move this to a common helper function?

integration-tests/braintree_test.go
```diff
@@ -31,30 +54,154 @@ func TestBraintree(t *testing.T) {
...
+	})
+	authRequest := sleet_testing.BaseAuthorizationRequest()
+	auth, err := client.Authorize(authRequest)
+	if err != nil {
+		t.Error("Authorize request should not have failed")
+	}
+
+	if auth.Success == false {
+		t.Error("Resulting auth should have been successful")
+	}
```
darshit-bolt, line 141: I fee like we don't need to have these checks everywhere since we already have a test which checks for valid auth for `sleet_testing.BaseAuthorizationRequest`

integration-tests/braintree_test.go
```diff
@@ -31,30 +54,154 @@ func TestBraintree(t *testing.T) {
...
+	if err != nil {
+		t.Error("Capture request should not have failed")
+	}
+
+	if capture.Success == false {
+		t.Error("Resulting capture should have been successful")
+	}
+}
+
+// TestBraintreeAuthPartialCapture
```
darshit-bolt, line 89: Should we add a test for PartialRefunds as well?

gateways/braintree/request_builders.go
nirajjayantbolt: No like this is specifically to the braintree.Decimal class - not general decimals?

darshit-bolt: Aah, I see, my bad

integration-tests/braintree_test.go
darshit-bolt: Wait, this says if `capture.Success is false` throw error. Seems correct to me @jakechiang-bolt

integration-tests/braintree_test.go
nirajjayantbolt: Yeah I dont know I kind of looked around at other examples and I kind of like the pattern of continually building checks. Even if its redundant, if I don't have those checks and make assumptions that the auth was successful then what if capture fails down below and I don't know why?

integration-tests/braintree_test.go
nirajjayantbolt: Eh haha might be overkill? But I don't mind

integration-tests/braintree_test.go
darshit-bolt: Ya makes sense. This pattern works!

integration-tests/braintree_test.go
jakechiang-bolt: Oh I just meant it as a stylistic change, switching `if capture.Success == false` to `if !capture.Success`.

nirajjayantbolt: Oh yeah nice catch! I'll change them

gateways/braintree/request_builders.go
```diff
@@ -3,27 +3,47 @@ package braintree
...
+		return nil, err
+	}
+
+	request := &braintree_go.TransactionRequest{
+		Type:   "sale",
+		Amount: amount,
+		CreditCard: &braintree_go.CreditCard{
 			Number:         card.Number,
 			ExpirationDate: fmt.Sprintf("%02d/%02d", card.ExpirationMonth, card.ExpirationMonth%100),
 			CVV:            card.CVV,
```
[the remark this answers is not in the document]
nirajjayantbolt: Oh good catch - let me take another look

integration-tests/braintree_test.go
darshit-bolt: Oh I see, haha, ya sounds good

gateways/braintree/request_builders.go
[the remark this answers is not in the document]
nirajjayantbolt: You're right! Great catch!

PR #83, opened 2020-05-21
[Misc] - add sleet environments to all PsPs

nirajjayantbolt:


nirajjayantbolt:
Ugh there are still more things I want to do:
1.) Why did I make Braintree require a separate credentials parameter?
2.) I should actually probably just generate the clients if they are the same for Stripe, BT, Adyen and store that so I don't replicate the same code 4 times for each call-site?

I'll do those in subsequent PRs

gateways/authorizenet/environment.go
```diff
@@ -0,0 +1,10 @@
+package authorizenet
+
+import "github.com/BoltApp/sleet/common"
+
+func authorizeNetHost(env common.Environment) string {
+	if env == common.Production {
+		return "https://api.authorize.net/xml/v1/request.api"
```
jakechiang-bolt, line 7: I'm not too familiar with Auth.net's API, are all requests sent to this URL path? If the path is always the same let's change the client to store `url` instead of `host` and update this to be `authorizeNetURL`.

nirajjayantbolt: makes sense and yeah it appears to be like this is used for all requests

gateways/authorizenet/authorizenet.go
```diff
@@ -10,28 +10,26 @@ import (
...
 	merchantName   string
 	transactionKey string
 	httpClient     *http.Client
+	host string
 }
 
 // NewClient uses authentication above with a default http client
-func NewClient(merchantName string, transactionKey string) *AuthorizeNetClient {
-	return NewWithHttpClient(merchantName, transactionKey, common.DefaultHttpClient())
+func NewClient(merchantName string, transactionKey string, environment common.Environment) *AuthorizeNetClient {
```
jakechiang-bolt, line 22: We should standardize whether env is the first or last arg to the `NewClient` functions since we do both right now. Any preference?

nirajjayantbolt: Ouch yeah thanks for the catch. Lets make it last except for HTTP client?

jakechiang-bolt: Sounds good.

nirajjayantbolt:
I mean, the braintreeClient we create right before we make the authorize call. Once we have the information from the caller we don't need to keep doing the whole btClient := newBraintreeClient. We should just store that client during the initialization call

jakechiang-bolt:
Ah, got it. Yeah that's a good idea.

PR #91, opened 2020-06-26
[NMI] Capture

dbook13:
Part 1/3 of [FP-209]. This PR adds support and a test for NMI captures. The NMI Request type should now have everything it needs for refunds and voids as well.

[FP-209]: https://boltpay.atlassian.net/browse/FP-209


dbook13:
Ya I was just following cybs, but good call I'll add in some more tests.

PR #92, opened 2020-06-29
[NMI] Void

dbook13:
Part 2/3 of [FP-209].

This PR adds support and tests for NMI voids. It also modifies the NMI tests to make it harder to make some of the mistakes I've run into writing them so far.

[FP-209]: https://boltpay.atlassian.net/browse/FP-209

### DIFF
--- a/gateways/nmi/request_builders.go
+++ b/gateways/nmi/request_builders.go
@@ -6,6 +6,13 @@ import (
 	"strconv"
 )
 
+// NMI transaction types
+const (
+	auth    = "auth"
+	capture = "capture"
+	void    = "void"
+)
+
 func buildAuthRequest(testMode bool, securityKey string, request *sleet.AuthorizationRequest) *Request {
 	zeroPad := ""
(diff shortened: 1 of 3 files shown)


gateways/nmi/request_builders.go
```diff
@@ -47,6 +47,15 @@ func buildCaptureRequest(testMode bool, securityKey string, request *sleet.Captu
 	}
 }
 
+func buildVoidRequest(testMode bool, securityKey string, request *sleet.VoidRequest) *Request {
+	return &Request{
+		SecurityKey:     securityKey,
+		TestMode:        enableTestMode(testMode),
+		TransactionID:   &request.TransactionReference,
+		TransactionType: "void",
```
nirajjayantbolt, line 55: nit: maybe consts at the top for this, auth, capture? Just to so we know where these values come from?

dbook13: Good call, moved

PR #97, opened 2020-07-13
Implement gateway for firstdata api

Steams:


gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+	apiSecret       string
+	clientRequestID string
+	httpClient      *http.Client
+}
+
+func NewClient(env common.Environment, apiKey, apiSecret string) *FirstdataClient {
+	return NewWithHttpClient(env, apiKey, apiSecret, common.DefaultHttpClient())
+}
+
+func NewWithHttpClient(env common.Environment, apiKey, apiSecret string, httpClient *http.Client) *FirstdataClient {
```
nirajjayantbolt, line 35: if this line wraps over can we do something like this:
```
func NewWithHttpClient(
env common.Environment, 
apiKey, 
apiSecret string, 
httpClient *http.Client,
) *FirstdataClient {

File: gateways/firstdata/firstdata.go [hunk 2]
```diff
@@ -0,0 +1,257 @@
...
+	url := "https://" + client.host + authPath
+
+	firstdataResponse, err := client.sendRequest(*clientRequestId, url, *firstdataAuthRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	success := false
+
+	if firstdataResponse.TransactionStatus == "APPROVED" || firstdataResponse.TransactionStatus == "WAITING" {
```
nirajjayantbolt, line 61: can we move these consts into our types file?

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+		host:       firstdataHost(env),
+		apiKey:     apiKey,
+		httpClient: httpClient,
+	}
+}
+
+func (client *FirstdataClient) Authorize(request *sleet.AuthorizationRequest) (*sleet.AuthorizationResponse, error) {
+	firstdataAuthRequest, err := buildAuthRequest(request)
+
+	clientRequestId := request.ClientTransactionReference
```
nirajjayantbolt, line 46: Why extract this out here? We can just use it from firstdataAuthRequest?

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	url := "https://" + client.host + authPath
```
nirajjayantbolt, line 52: even this url building can go into sendRequest

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, err
+	}
+
+	success := false
+
+	if firstdataResponse.TransactionStatus == "APPROVED" || firstdataResponse.TransactionStatus == "WAITING" {
+		success = true
+	}
+
+	if firstdataResponse.Error != nil {
```
nirajjayantbolt, line 65: lets check the error before the transaction status

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+		response := sleet.AuthorizationResponse{Success: false, ErrorCode: firstdataResponse.Error.Code}
+		return &response, nil
+	}
+
+	avsRaw, err := json.Marshal(firstdataResponse.Processor.AVSResponse)
```
nirajjayantbolt, line 70: This should already be marshalled right? Why not just call AvsRawString = string(firstdataResponse.Processor.AVSResponse?)

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,257 @@
...
+
+	if firstdataResponse.Error != nil {
+		response := sleet.RefundResponse{Success: false, ErrorCode: &firstdataResponse.Error.Code}
+		return &response, nil
+	}
+	return &sleet.RefundResponse{Success: true, TransactionReference: firstdataResponse.IPGTransactionId}, nil
+}
+
+// reqId is our internally generated id,unique per request, tranasactionRef is firstdata's returned ref
+func (client *FirstdataClient) TransactionInquiry(reqId, transactionRef string) (*Response, error) {
```
nirajjayantbolt, line 149: what is this being used for?

gateways/firstdata/request_builders.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/BoltApp/sleet"
+)
+
+func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error) {
+
+	amountStr := sleet.AmountToString(&authRequest.Amount)
+
+	request := &Request{
+		RequestType: "PaymentCardPreAuthTransaction",
```
nirajjayantbolt, line 14: lets put const in types file

gateways/firstdata/request_builders.go
```diff
@@ -0,0 +1,62 @@
...
+			Total:    amountStr,
+			Currency: authRequest.Amount.Currency,
+		},
+		PaymentMethod: PaymentMethod{
+			PaymentCard: PaymentCard{
+				Number:       authRequest.CreditCard.Number,
+				SecurityCode: authRequest.CreditCard.CVV,
+				ExpiryDate: ExpiryDate{
+					Month: strconv.Itoa(authRequest.CreditCard.ExpirationMonth),
+					Year:  strconv.Itoa(authRequest.CreditCard.ExpirationYear)[2:],
```
nirajjayantbolt, line 25: lets add comment explaining we only use 2 digit year? Maybe also make this a little more robust (like a separate function we can also unit test)

gateways/firstdata/translation.go
```diff
@@ -0,0 +1,40 @@
+package firstdata
+
+import "github.com/BoltApp/sleet"
+
+// translateCvv converts a Firstdata CVV response code to its equivalent Sleet standard code.
+// TODO do case for NOT_CERTIFIED
+func translateCvv(rawCvv string) sleet.CVVResponse {
+	switch rawCvv {
+	case "MATCHED":
```
roopakv, line 9: what if we simply had a map we could get value from map and if not present return unknown? easier to read

```
var firstdataCvvMap = map[string]sleet.CVVResponse{
  "MATCHED": sleet.CVVResponseMatch,
  "NOT_MATCHED": sleet.CVVResponseNoMatch,
  ...
}

cvvCode, ok := firstdataCvvMap[rawCvv]
if !ok {
	return sleet.CVVResponseUnknown
}
return cvvCode
```

gateways/firstdata/firstdata.go
Steams: Subsequent requests require the transactionReference in the url but the initial auth request does not. So i built the url's outside of send and pass them n as opposed to passing an optional transactionReference variable to the send function

gateways/firstdata/firstdata.go
Steams: I used it in the integration tests to check the state of a transaction after secondary transactions. So after calling Capture i use inquiry to check if the state is "CAPTURED". It may seem unnecessary but the call to Capture does not actually return the state "CAPTURED" in its response, the inquiry does. 

Also, we can't assume that just because the call to Capture was successful, that the state must be "CAPTURED" because if the endpoint is called with a previously used client-request-id, it will automatically return the original response associated with that request id. So it is possible to get a successful http response from calling capture, but not actually capture the transaction.

gateways/firstdata/firstdata.go
nirajjayantbolt: Okay sure got it - I realized we did something similar for Cybersource so that's okay. Can we put URLs to the capture/void/refund just so we know how we construct the URLs?

gateways/firstdata/firstdata.go
nirajjayantbolt: If its just used for integration tests can we move the function into the test file? I'd like to kee the actual gateway file just for like production processing, if that makes sense?

gateways/firstdata/translation.go
nirajjayantbolt: Maps are great, I should actually update other gateways to do this...

gateways/firstdata/firstdata.go
Steams: AVSResponse is a struct. string(AVSResponse) would throw a cannot convert error. I could use fmt.Sprintf("%v",AVSResponse) if you would prefer that.

nirajjayantbolt: Wait I guess I meant, when you unmarshalled the response into the firstdataResponse, AVS was also unmarshalled right?

gateways/firstdata/types.go
```diff
@@ -0,0 +1,103 @@
...
+	ApiTraceID      string `json:"apiTraceId"`
+	ResponseType    string `json:"responseType"`
+	Ipg             string `json:"responseType"`
+}
+type Detail struct {
+	Field   string `json:"field"`
+	Message string `json:"message"`
+}
+
+type TransactionAmount struct {
```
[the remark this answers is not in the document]
Steams: There are 2 "Amount" structs "TransactionAmount" and "ApprovedAmount". One takes a string value the other takes a float value. Should i still change TransactionAmount to Amount ?

gateways/firstdata/firstdata.go
Steams: Yes, it was UnMarshalled from a Json string into a Go struct, at the end of the send function, which then returns that Go struct. It can't (as far as i know, i tried your suggestion just now and it was an error) be automatically converted back into a string without marshaling it back to a json string with json.Marshal().

```
type ProcessorData struct {
        ...
	AVSResponse             AVSResponse `json:"avsResponse"`
}

type AVSResponse struct {
	StreetMatch   string `json:"streetMatch"`  
	PostCodeMatch string `json:"postCodeMatch"` 
}

```

gateways/firstdata/firstdata.go
Steams: Sure thing

gateways/firstdata/types.go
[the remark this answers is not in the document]
daisy1754: ouch I missed that ok let's keep TransactionAmount then

gateways/firstdata/request_builders.go
Steams: Added the comment in the types file. I havent pushed the unit test code yet but it should be covered by the unit test for the requestBuilder. Can discuss again after that push

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,215 @@
...
+
+type FirstdataClient struct {
+	host            string
+	apiKey          string
+	apiSecret       string
+	clientRequestID string
+	httpClient      *http.Client
+}
+
+func NewClient(env common.Environment, apiKey, apiSecret string) *FirstdataClient {
```
nirajjayantbolt, line 30: Can we actually take in a new struct called `Credentials` that uses apiKey, apiSecret as a string? I have since learned that this is a better pattern because if we have to add variables later on we don't have to change the function signature

gateways/firstdata/firstdata.go
nirajjayantbolt: So at the end `avsRaw` will be something like `{"streetMatch": "Y", "postCodeMatch": "Y"}`? I think it might be cleaner to actually just do something like fmt.Sprintf("%s:%s", avsResponse.StreetMatch, avsResponse.PostCodeMatch) so we don't put the "streetMatch" and "postCodeMatch" strings in there

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,215 @@
...
+	httpClient *http.Client,
+) *FirstdataClient {
+	return &FirstdataClient{
+		host:       firstdataHost(env),
+		apiKey:     apiKey,
+		httpClient: httpClient,
+	}
+}
+
+func (client *FirstdataClient) primaryURL() string {
```
nirajjayantbolt, line 47: nit: can we put comments at the top of all of these functions explaining what they will be used for

gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,215 @@
...
+func (client *FirstdataClient) sendRequest(reqId, url string, data Request) (*Response, error) {
+
+	bodyJSON, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+
+	timestamp := strconv.FormatInt(time.Now().Unix(), 10)
+
+	hashData := client.apiKey + reqId + timestamp + string(bodyJSON)
```
nirajjayantbolt, line 177: can we paste a URL showing the requirements for singing the body

Steams: Place the link as a comment in the code or just here ?

nirajjayantbolt: Paste a comment as the link in code

PR #98, opened 2020-07-21
Add structs for marshaling Firstdata request/response json

Steams:
The docs dont support linking directly to the relevant models. https://docs.firstdata.com/org/gateway/docs/api#models
We're looking at

- PrimaryTransaction - https://docs.firstdata.com/org/gateway/docs/api#create-primary-transaction
- SecondaryTransaction - https://docs.firstdata.com/org/gateway/docs/api#secondary-transaction
- PaymentCardPreAuthTransaction
- PostAuthTransaction
- ReturnTransaction
- VoidTransaction
- Amount
- PaymentCardPaymentMethod
- TransactionStatus
- TransactionState
- ProcessorData
- Error
- AVSResponse
- securityCodeResponse

### DIFF
--- a/gateways/firstdata/types.go
+++ b/gateways/firstdata/types.go
@@ -1,0 +1,149 @@
+package firstdata
+
+// RequestType represents the valid requestType values that can be sent in a firstdata request
+type RequestType string
+
+const (
+	RequestTypeAuth    RequestType = "PaymentCardPreAuthTransaction"
+	RequestTypeCapture RequestType = "PostAuthTransaction"
+	RequestTypeRefund  RequestType = "ReturnTransaction"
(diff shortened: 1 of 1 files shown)


gateways/firstdata/types.go
```diff
@@ -0,0 +1,149 @@
+package firstdata
+
+// RequestType represents the valid requestType values that can be sent in a firstdata request
+type RequestType string
+
+const (
+	RequestTypeAuth    RequestType = "PaymentCardPreAuthTransaction"
```
nirajjayantbolt, line 7: just checking - sale transaction means no capture, right?

Steams: Yeah, PreAuth first, then PostAuth to capture

PR #99, opened 2020-07-22
Request builders and avs/cvv response code translators and unit tests

Steams:


gateways/firstdata/request_builders.go
```diff
@@ -0,0 +1,62 @@
+package firstdata
+
+import (
+	"strconv"
+
+	"github.com/BoltApp/sleet"
+)
+
+func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error) {
+
```
nirajjayantbolt, line 10: nit: can we remove this line?

gateways/firstdata/request_builders.go
```diff
@@ -0,0 +1,62 @@
+package firstdata
+
+import (
+	"strconv"
+
+	"github.com/BoltApp/sleet"
+)
+
+func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error) {
```
nirajjayantbolt, line 9: Also, not sure if this is because I mistakenly did this in other areas but if there is no error lets just return back *Request

PR #100, opened 2020-07-23
Utilities functions for sending firstdata requests and associated tests

Steams:


gateways/firstdata/firstdata.go
```diff
@@ -0,0 +1,109 @@
...
+	httpClient      *http.Client
+}
+
+// Credentials contains the merchant api key and secret for the firstdata gateway
+type Credentials struct {
+	ApiKey    string
+	ApiSecret string
+}
+
+// NewClient creates a new firstdataClient with the given credentials and an httpClient
```
nirajjayantbolt, line 35: nit: technically this is using the default client for the comment?

PR #101, opened 2020-07-23
Authorize and Capture methods

Steams:


Steams:
> This is great! I am still having trouble getting a test account with them set up so I feel like unit tests are good here. Do you want to also add integration tests in a separate PR?

Sure, I'll put up a PR with integration tests for auth and capture next

PR #102, opened 2020-07-24
Integration tests for auth and capture

Steams:


integration-tests/firstdata_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+	if err != nil {
+		t.Fatalf("Got runtime error while running capture %q", err)
+	}
+
+	if capResp.ErrorCode != nil {
+		t.Errorf("Expected No Error Code: received: %s", *capResp.ErrorCode)
+	}
+
+	inq, err := transactionInquiry(uuid.New().String(), capResp.TransactionReference)
```
nirajjayantbolt, line 63: why dont we use the captureRequest UUID here?

nirajjayantbolt:
Oh also for this to work in CircleCI I will need to set up the env variables. Looking at that this morning

integration-tests/firstdata_test.go
Steams: The first uuid param here is supposed to be newly generated for every request, used for their internal tracking. The second parameter is the reference to the transaction we are querying.

PR #105, opened 2020-08-07
Type definitions

Steams:
There is a very long list of currency codes and currency exponents for each country in the doc. If there is a specific subset of these you are interested I can pull them out into defined types

### DIFF
--- a/gateways/orbital/types.go
+++ b/gateways/orbital/types.go
@@ -94,7 +94,7 @@ type Response struct {
 }
 
 type RequestBody struct {
-	XMLName                   xml.Name      //set dynamically : NewOrder,Capture,Reversal
+	XMLName                   xml.Name
 	OrbitalConnectionUsername string        `xml:"OrbitalConnectionUsername"`
 	OrbitalConnectionPassword string        `xml:"OrbitalConnectionPassword"`
 	BIN                       BIN           `xml:"BIN"`
@@ -118,12 +118,12 @@ type RequestBody struct {
 	AVSstate                  string        `xml:"AVSstate,omitempty"`
 	AVScity                   string        `xml:"AVScity,omitempty"`
 	AVSname                   string        `xml:"AVSname,omitempty"`
-	AVScountryCode            string        `xml:"AVScountryCode,omitempty"` // TODO verify spelling
+	AVScountryCode            string        `xml:"AVScountryCode,omitempty"`
 	AVSphoneNum               string        `xml:"AVSphoneNum,omitempty"`
 }
 
 type ResponseBody struct {
-	XMLName        xml.Name        //set dynamically : NewOrder,Capture,Reversal
+	XMLName        xml.Name
 	IndustryType   string          `xml:"IndustryType"`
 	MessageType    string          `xml:"MessageType"`
 	MerchantID     int             `xml:"MerchantID"`
--- a/gateways/orbital/types.go
+++ b/gateways/orbital/types.go
@@ -5,10 +5,9 @@ import "encoding/xml"
 type RequestType string
 
 const (
-	RequestTypeAuth    = "NewOrder"
-	RequestTypeCapture = "MarkForCapture"
-	RequestTypeVoid    = "Reversal"
-	RequestTypeRefund  = "Reversal"
+	RequestTypeNewOrder = "NewOrder"
+	RequestTypeCapture  = "MarkForCapture"
+	RequestTypeVoid     = "Reversal"
 )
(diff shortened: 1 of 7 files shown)


gateways/orbital/types.go
```diff
@@ -0,0 +1,127 @@
...
+
+import "encoding/xml"
+
+type RequestType string
+
+const (
+	RequestTypeAuth    = "NewOrder"
+	RequestTypeCapture = "MarkForCapture"
+	RequestTypeVoid    = "Reversal"
+	RequestTypeRefund  = "Reversal"
```
nirajjayantbolt, line 11: No need for both

gateways/orbital/types.go
```diff
@@ -0,0 +1,127 @@
...
+	AVSResponseNotChecked             AVSResponseCode = "3"
+	AVSResponseSkipped4               AVSResponseCode = "4"
+	AVSResponseSkippedR               AVSResponseCode = "R"
+	AVSResponseMatch                  AVSResponseCode = "H"
+	AVSResponseNoMatch                AVSResponseCode = "G"
+	AVSResponseZipMatchAddressNoMatch AVSResponseCode = "A"
+	AVSResponseZipNoMatchAddressMatch AVSResponseCode = "F"
+)
+
+type Request struct {
```
nirajjayantbolt, line 73: these seem a little odd to me, why XML Name?

gateways/orbital/types.go
```diff
@@ -0,0 +1,127 @@
...
+type RequestBody struct {
+	XMLName                   xml.Name
+	OrbitalConnectionUsername string        `xml:"OrbitalConnectionUsername"`
+	OrbitalConnectionPassword string        `xml:"OrbitalConnectionPassword"`
+	BIN                       BIN           `xml:"BIN"`
+	TerminalID                string        `xml:"TerminalID"` // usually 001, for PNS can be 001 - 999 but usually 001
+	IndustryType              IndustryType  `xml:"IndustryType,omitempty"`
+	MessageType               MessageType   `xml:"MessageType,omitempty"`
+	MerchantID                int           `xml:"MerchantID,omitempty"`
+	AccountNum                string        `xml:"AccountNum,omitempty"`
```
nirajjayantbolt, line 92: is this cc number?

gateways/orbital/types.go
Steams: All requests can use the same structure and properties but the name of the actual tag when marshaled to XML has to change, xmlName lets you change it. EG :

an auth request looks like:
```
<Request> //This is the Request Struct
  <NewOrder>...</NewOrder> //this is a RequestBody struct with xmlName = "NewOrder"
</Request>
```

a void request looks like:
```
<Request> //This is the Request Struct
  <Reversal>...</Reversal> //this is a RequestBody struct with xmlName = "Reversal"
</Request>
```

Setting it looks like this :
`orbitalAuthRequest.Body.XMLName = xml.Name{Local: RequestTypeAuth}`

gateways/orbital/types.go
Steams: Yes

gateways/orbital/types.go
nirajjayantbolt: Thanks got it

gateways/orbital/types.go
```diff
@@ -0,0 +1,150 @@
...
+	RequestTypeCapture  = "MarkForCapture"
+	RequestTypeVoid     = "Reversal"
+)
+
+type MessageType string
+
+const (
+	MessageTypeAuth           MessageType = "A"
+	MessageTypeAuthAndCapture MessageType = "AC"
+	MessageTypeCapture        MessageType = "FC"
```
dbook13, line 18: Can we call this `MessageTypeForceAndCapture`? It seems like there is a slight difference between a Force and Capture and a normal capture, and I want to make sure they don't get confused here

Steams: Sure thing, just to be clear though, the Capture method i implemented actually uses the "MarkForCapture" request type, as opposed to a NewOrder request with the AuthAndCApture or ForceCapture message types. 
Similar to reversal vs refund earlier, is it that the method of capturing we're interested in should use the NewOrder request and not MarkForCapture ?

dbook13: We want to use the "MarkForCapture" request type for the Capture method like you've already implemented. I just wanted the naming of this message type to make it clear that it is not a secondary method of performing a transaction and actually does something different.

PR #106, opened 2020-08-07
[NMI] exp date for one off test transactions

dbook13:
In order to send a one off test transaction to a merchant account that isn't in test mode, we need to be using one of the test cards and the expiration date defined here: https://secure.networkmerchants.com/gw/merchants/resources/integration/integration_portal.php#testing_information


dbook13:
Can't change the expiration date here because it breaks Adyen, so I'm going to add it in hail instead

PR #109, opened 2020-08-24
request builders and translations

Steams:


gateways/orbital/request_builders.go
```diff
@@ -0,0 +1,84 @@
...
+	body := RequestBody{
+		IndustryType:     IndustryTypeEcomm,
+		MessageType:      MessageTypeAuth,
+		BIN:              BINStratus,
+		TerminalID:       TerminalIDStratus,
+		AccountNum:       authRequest.CreditCard.Number,
+		Exp:              exp,
+		CurrencyCode:     code,
+		CurrencyExponent: CurrencyExponentDefault,
+		CardSecValInd:    CardSecPresent,
```
dbook13, line 25: From the NewOrder message definition it seems like this value should be null if the transaction is not a Visa or Discover transaction. I'm not sure if we have a way to do this currently, so I think this is fine for now but I just want to make note of it

gateways/orbital/request_builders.go
```diff
@@ -0,0 +1,84 @@
...
+
+	amount := authRequest.Amount.Amount
+	exp := strconv.Itoa(authRequest.CreditCard.ExpirationYear) + strconv.Itoa(authRequest.CreditCard.ExpirationMonth)
+	code := currencyMap[authRequest.Amount.Currency]
+
+	body := RequestBody{
+		IndustryType:     IndustryTypeEcomm,
+		MessageType:      MessageTypeAuth,
+		BIN:              BINStratus,
+		TerminalID:       TerminalIDStratus,
```
dbook13, line 20: Are we just assuming BIN and Terminal ID are Stratus for simplicity sake currently?

Steams: Yes, not sure what the logic would be for determining when to use "Stratus" vs "Tandem"

dbook13: Ya I think it would have to be an additional value we accept. I think most merchants for card-not-present will be Stratus, so we can just worry about it if there's ever a merchant that isn't Stratus

PR #111, opened 2020-08-26
Auth and capture

Steams:


gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	authRequest.Body.OrbitalConnectionPassword = client.credentials.password
+	authRequest.Body.MerchantID = client.credentials.merchantID
+
+	orbitalResponse, err := client.sendRequest(authRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	if orbitalResponse.Body.ProcStatus != 0 {
+		response := sleet.AuthorizationResponse{Success: false, ErrorCode: strconv.Itoa(orbitalResponse.Body.ProcStatus)}
```
dbook13, line 47: nit: define error code above like you do in `Capture`

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	reader := bytes.NewReader(bodyXML)
+	request, err := http.NewRequest(http.MethodPost, client.host, reader)
+	if err != nil {
+		return nil, err
+	}
+	request.Header.Add("Content-Type", "application/xml; charset=utf-8")
```
dbook13, line 100: It looks like there are other mandatory headers from the MIME header section of the orbital specification (page 15). Do we need to add these as well?

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	merchantID int
+}
+
+type OrbitalClient struct {
+	host        string
+	credentials Credentials
+	httpClient  *http.Client
+}
+
+func NewClient(env common.Environment, credentials Credentials) *OrbitalClient {
```
dbook13, line 26: Are you planning on adding a `NewWithHttpClient` in a later PR? It seems like `NewClient` functions in most other gateways create a default http client and then call `NewWithHttpClient`

nirajjayantbolt: +1 can we do the NewWithHttpClient?

gateways/orbital/environment.go
```diff
@@ -0,0 +1,12 @@
+package orbital
+
+import (
+	"github.com/BoltApp/sleet/common"
+)
+
+func orbitalHost(env common.Environment) string {
+	if env == common.Production {
+		return "https://orbital1.chasepaymentech.com/authorize"
```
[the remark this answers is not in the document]
nirajjayantbolt: Yeah we can even make a random number generator if we want but lets not worry about that yet

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	return &OrbitalClient{
+		host:        orbitalHost(env),
+		credentials: credentials,
+		httpClient:  common.DefaultHttpClient(),
+	}
+}
+
+func (client *OrbitalClient) Authorize(request *sleet.AuthorizationRequest) (*sleet.AuthorizationResponse, error) {
+
+	authRequest := buildAuthRequest(request)
```
nirajjayantbolt, line 36: lets just send the credentials down into buildAuthRequest so we can do it there?

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	"io/ioutil"
+	"net/http"
+	"strconv"
+
+	"github.com/BoltApp/sleet"
+	"github.com/BoltApp/sleet/common"
+)
+
+type Credentials struct {
+	username   string
```
nirajjayantbolt, line 15: Make all of these public

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,122 @@
...
+	password   string
+	merchantID int
```
nirajjayantbolt, line 17: string not int

nirajjayantbolt: Wait actually nevermind, didnt see that XML takes an int value

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (client *OrbitalClient) sendRequest(data Request) (*Response, error) {
+
+	bodyXML, err := xml.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+
+	reader := bytes.NewReader(bodyXML)
```
dbook13, line 125:
```suggestion
	bodyWithHeader := xml.Header + string(bodyXML)
	reader := bytes.NewReader([]byte(bodyWithHeader))
```

gateways/orbital/orbital.go
```diff
@@ -0,0 +1,159 @@
...
+
+	bodyWithHeader := xml.Header + string(bodyXML)
+	reader := bytes.NewReader([]byte(bodyWithHeader))
+	request, err := http.NewRequest(http.MethodPost, client.host, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	request.Header.Add("MIME-Version", "1.1")
+	request.Header.Add("Content-Type", "application/PTI80")
```
nirajjayantbolt, line 133: maybe we make this a const so its more clear where this value comes from? And MIME Version?

gateways/orbital/orbital_test.go
```diff
@@ -0,0 +1,245 @@
...
+	request.Body.OrbitalConnectionUsername = "username"
+	request.Body.OrbitalConnectionPassword = "password"
+	request.Body.MerchantID = 1
+
+	t.Run("With Successful Response", func(t *testing.T) {
+		httpmock.Activate()
+		defer httpmock.DeactivateAndReset()
+
+		httpmock.RegisterResponder("POST", url, func(req *http.Request) (*http.Response, error) {
+			headerReceived = req.Header
```
nirajjayantbolt, line 81: lets test headers too - we can do in a separate PR?

gateways/orbital/request_builders.go
```diff
@@ -51,10 +55,13 @@ func buildCaptureRequest(captureRequest *sleet.CaptureRequest) Request {
 	}
 
 	body.XMLName = xml.Name{Local: RequestTypeCapture}
+	body.OrbitalConnectionUsername = credentials.Username
```
nirajjayantbolt, line 58: wait why dont we just move this up into the initialization of the struct?

gateways/orbital/orbital.go
dbook13: Made all the headers consts

gateways/orbital/orbital_test.go
dbook13: sounds good

gateways/orbital/request_builders.go
dbook13: moved

PR #113, opened 2020-08-31
Change Ordering of Fields

nirajjayantbolt:
I've noticed some problems in postman if ordering is weird so fixing some issues I saw and adding some ApplePay stuff

### DIFF
--- a/gateways/orbital/types.go
+++ b/gateways/orbital/types.go
@@ -121,29 +121,29 @@ type RequestBody struct {
 	XMLName                   xml.Name
 	OrbitalConnectionUsername string           `xml:"OrbitalConnectionUsername"`
 	OrbitalConnectionPassword string           `xml:"OrbitalConnectionPassword"`
-	BIN                       BIN              `xml:"BIN"`
-	TerminalID                string           `xml:"TerminalID"` // usually 001, for PNS can be 001 - 999 but usually 001
 	IndustryType              IndustryType     `xml:"IndustryType,omitempty"`
 	MessageType               MessageType      `xml:"MessageType,omitempty"`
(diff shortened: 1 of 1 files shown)


gateways/orbital/types.go
```diff
@@ -121,29 +121,31 @@ type RequestBody struct {
...
 	AVSaddress2               *string          `xml:"AVSaddress2,omitempty"`
-	AVSstate                  string           `xml:"AVSstate,omitempty"`
 	AVScity                   string           `xml:"AVScity,omitempty"`
+	AVSstate                  string           `xml:"AVSstate,omitempty"`
 	AVSname                   string           `xml:"AVSname,omitempty"`
 	AVScountryCode            string           `xml:"AVScountryCode,omitempty"`
 	AVSphoneNum               string           `xml:"AVSphoneNum,omitempty"`
+	OrderID                   string           `xml:"OrderID,omitempty"`                // generated id, max 22 chars
+	Amount                    int64            `xml:"Amount,omitempty"`                 //int with the last 2 digits being implied decimals ie 100.25 is sent as 10025, 90 is sent as 9000
+	DPANInd                   string           `xml:"DPANInd,omitempty"`                // does this token represent a device based Primary Account Number (DPAN)
```
dbook13, line 147: From the orbital docs (page 131), it seems like this value is just a single letter, `Y`, when we are submitting new DPAN auths. I'm a bit confused by the comment here, it kind of makes it seem like this is where we actually put the DPAN even though I thin DPAN is supposed to go in AccountNum

nirajjayantbolt: Ah good call out, I'll update comment. Its an indicator just like Y meaning we are using DPAN

PR #115, opened 2020-09-02
void and refund

Steams:
note that void responses do not ever contain a RespCode


gateways/orbital/types.go
```diff
@@ -121,29 +121,29 @@ type RequestBody struct {
 	XMLName                   xml.Name
 	OrbitalConnectionUsername string           `xml:"OrbitalConnectionUsername"`
 	OrbitalConnectionPassword string           `xml:"OrbitalConnectionPassword"`
+	BIN                       BIN              `xml:"BIN"`
```
dbook13, line 124: Why did you reorder the variables in these structs?

dbook13 (approved):

gateways/orbital/types.go
Steams: In the unit tests a match the generated request body against an existing "valid" request body in the test_data folder. Changing the order in the struct definitions changes the order of the generated request body making it not match the existing sample request (because the comparison between the 2 is just a string comparison). A bit volatile yes, perhaps there is a way to compare 2 XML structs by value instead of by comparing there strings representations.

dbook13: Could you use the existing order and change the order of fields in the sample requests/responses? Niraj noticed that Orbital is super picky about the ordering of fields matching the ordering of fields in the Message Definition in the docs.

dbook13: Thanks!

PR #116, opened 2020-09-04
Integration tests

Steams:


nirajjayantbolt (commented):
Unfortunately we are going to want to stop here as well, I don't want to use our merchants credentials in our circleci pipeline so we have to wait until Bolt itself is certified

PR #118, opened 2020-09-08
[nmi] return response text for declines

dbook13:
Transaction references are sometimes returned in the response text when the transaction is declined. This reference lets merchants reach out to NMI to get more information about the decline.


nirajjayantbolt (approved):

dbook13:
Just going to turn off response sanitization in hail instead

PR #120, opened 2020-09-11
Initiator flag types

nirajjayant:


dbook13 (approved):
Copying over seems fine to me. Should we eventually replace the type declaration in hail and just replace it by importing the type from sleet?

PR #123, opened 2020-09-14
[adyen] ContAuth shopperInteraction

dbook13:
Add ContAuth shopper interaction to adyen credit card auth requests that will allow subsequent requests with that credit card to be authed without a cvv.

Confirmed that this change fixes the loggedIn and recharge auth flows.


nirajjayant (approved):
Works for me - I suppose we could send this with CVV once but I don't think it matters

dbook13:
> Works for me - I suppose we could send this with CVV once but I don't think it matters

As it is currently we do send the request with CVV for any credit cards that aren't already in our system, but auths for a card we have saved in our system are sent without CVV

nirajjayant:
Yeah I verified with Postman that it wont matter

PR #124, opened 2020-09-15
[adyen] new vs. existing customer ShopperInteraction

dbook13:
Modify Adyen request (ShopperInteraction specifically) based on if CVV is present in the sleet request which is a proxy for if the customer is new / a guest or an existing customer with a saved card.


nirajjayant (approved):
Nice thanks for adding the test

I think we should make a note about this maybe at the Adyen gateway level just so our callers know they must be compliant? Wdyt?

dbook13:
> I think we should make a note about this maybe at the Adyen gateway level just so our callers know they must be compliant? Wdyt?

Ya I'll add a note above the Authorize function

PR #125, opened 2020-09-18
[adyen] saving new customer credit card information

dbook13:
In the Adyen auth request, differentiate between new customers who do want to save their credit card information and new customer who do not want to save their credit card information.


nirajjayant (approved):
Nit: add one comment on Save in CC model so we know what its about

PR #129, opened 2020-09-24
[adyen] level3 data modifications

dbook13:
- Added list of unit of measurement codes
- Break up Adyen buildAuthRequest
- Omit optional level 3 fields from auth request if we don't have data for them


gateways/adyen/request_builder.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/adyen/adyen-go-api-library/v2/src/checkout"
 	"github.com/adyen/adyen-go-api-library/v2/src/payments"
 	"strconv"
+	"time"
+)
+
+const (
+	maxLineItemDescriptionLength = 26
+	timeFormatExample            = "020106"
```
nirajjayant, line 15: I believe this is a constant in the time package?

gateways/adyen/request_builder.go
```diff
@@ -130,3 +138,12 @@ func buildVoidRequest(voidRequest *sleet.VoidRequest, merchantAccount string) *p
 	}
 	return request
 }
+
+// NAifEmpty returns "NA" if the parameter is an empty string. Adyen best practices
+// for Level 2/3 data are to set the value of any fields that are unknown to "NA"
+func NAifEmpty(str string) string {
```
nirajjayant, line 144: can we make this a little more general? Like DefaultIfEmpty that takes a default value and returns that instead of NA? ANd we can move that to sleet common

utils.go
```diff
@@ -11,3 +11,12 @@ func AmountToString(amount *Amount) string {
 func AmountToDecimalString(amount *Amount) string {
 	return fmt.Sprintf("%.2f", float64(amount.Amount)/100.0)
 }
+
+// TruncateString returns a prefix of str of length truncateLength or all of
+// str if truncateLength is greater than len(str)
+func TruncateString(str string, truncateLength int) string {
```
nirajjayant, line 17: For our go-report card later can we also unit test this?

gateways/adyen/request_builder.go
dbook13: Done

gateways/adyen/request_builder.go
dbook13: Wasn't able to find a constant for this specific format but not using this anymore so not a concern

utils.go
dbook13: Added some tests

gateways/adyen/request_builder.go
```diff
@@ -130,3 +150,9 @@ func buildVoidRequest(voidRequest *sleet.VoidRequest, merchantAccount string) *p
 	}
 	return request
 }
+
+func addIfNonEmpty(value string, key string, data *map[string]string) {
```
dbook13, line 154: @niraj this seemed too specific to put in utils but what do you think?

units_of_measurement.go
```diff
@@ -0,0 +1,413 @@
+package sleet
```
dbook13, line 1: @nirajjayant we are starting to have a lot of files in this main sleet package, would it make sense to move some of them to `common`?

nirajjayant (approved):

units_of_measurement.go
nirajjayant: Yeah I think currency, string stuff should move to common or utils or something

gateways/adyen/request_builder.go
nirajjayant: Yeah lets keep it here for now but if we find ourselves using something like this again we can revisit?

PR #135, opened 2020-10-26
Fix authnet test and enable unit test in CI

daisy1754:
As far as I can tell unit test for auth.net is completely broken. Fixed them and update circleci config to run unit tests.

Note CI will be broken because integration test relied on secrets that are not accessible to non-Bolt employees :/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,10 @@ jobs:
           command: |
             go build -v ./...
       - run:
+          name: Unit Test
+          command: |
+            go test -v -tags=unit $(go list ./... | grep -v integration-tests)
+      - run:
           name: Integration Test
           command: |
             go test -v ./integration-tests/*.go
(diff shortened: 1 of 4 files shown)


nirajjayant:
@daisy1754 so I should move this env variables outside of the Bolt ecosystem?

daisy1754:
@nirajjayant looks like you may need to enable this setting - https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests

nirajjayant:
Awesome, looking

nirajjayant:
@daisy1754I chatted with Roopak and he doesnt want us to enable this in case we use contexts with Bolt's core repositories at some point. We added you as an external collaborator to Bolt, can we open the PR from the main repo not the fork?

daisy1754:
I'll try! I'll also send PR to disable integration test if env var is not available; otherwise external collaborator would have hard time

roopakv:
@daisy1754 added you as an external collaborator to the repo. if you close this PR and open one from the main repo directly things should work fine for now

PR #142, opened 2020-11-21
Convert switch statements in transaction.go to map

daisy1754:
so it's consistent with other file eg `authorizenet/transaction.go`. Also this helps increasing coverage :P because if we do switch we'd have to test all cases.


nirajjayant (approved):
Nice thanks for all of this

PR #149, opened 2020-12-05
Adyen - switch logic to only set successful if authorised, else treat…

nirajjayant:
… as failed

We found that there is a 3DS response and this logic is kind of backwards, we should only treat as success if we get authorised

### DIFF
--- a/gateways/adyen/adyen.go
+++ b/gateways/adyen/adyen.go
@@ -78,12 +78,12 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 		}
 	}
 
-	if result.ResultCode == adyen_common.Refused || result.ResultCode == adyen_common.Error {
+	if result.ResultCode == adyen_common.Authorised {
+		response.Success = true
+	} else {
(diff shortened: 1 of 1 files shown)


gateways/adyen/adyen.go
```diff
@@ -78,12 +78,12 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 		}
 	}
 
-	if result.ResultCode == adyen_common.Refused || result.ResultCode == adyen_common.Error {
+	if result.ResultCode == adyen_common.Authorised {
```
namanhkapur, line 81: not sure if it's worth, but we can write this without the else case, `if !result.ResultCode == adyen_common.Authorised`

namanhkapur (approved):
nice

gateways/adyen/adyen.go
nirajjayant: Yeah actually I hate if/else let me fix it

jakechiang-bolt (approved):

PR #152, opened 2020-12-18
[authnet] Add Last4 to RefundRequest

dbook13:
Per discussion [here](https://github.com/BoltApp/source/pull/10650#issuecomment-747713085), it no longer makes sense to use the refund request's `Options` field to store the last 4 digits of a credit card only for auth.net. This PR adds a new field to refund request, `Last4`, where we will now store this information.

### DIFF
--- a/gateways/authorizenet/authorizenet_test.go
+++ b/gateways/authorizenet/authorizenet_test.go
@@ -357,20 +357,6 @@ func TestRefund(t *testing.T) {
 		}
 	})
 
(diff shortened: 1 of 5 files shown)


nirajjayant (approved):
I think you can leave options in there for now even if we don't knowingly use it right now, but moving Last4 to being top level makes sense to me

dbook13:
> I think you can leave options in there for now even if we don't knowingly use it right now, but moving Last4 to being top level makes sense to me

Added Options back into both requests and updated the PR title and description.

PR #153, opened 2020-12-18
[authnet] return transaction ref for capture, void, and refund

dbook13:
Not returning a transaction ref for captures causes refunds to fail. Found this out while testing https://github.com/BoltApp/source/pull/10650


nirajjayant (approved):
Lol sorry I guess testing is the only way to find all of these

PR #154, opened 2021-01-04
add real time account updater types

namanhkapur:
- adds sleet wide RTAU constants
- adds adyen specific mapping from their representation to newly created sleet representation
- adds RTAU to Authorization Response struct


types.go
```diff
@@ -153,3 +157,14 @@ type Currency struct {
...
+
+// RTAUStatus represents the Real Time Account Updater response from a processor, if applicable
+type RTAUStatus string
+
+const (
+	RTAUStatusUnknown      RTAUStatus = "Unknown"
+	RTAUStatusCardChanged  RTAUStatus = "CardChanged"
+	RTAUStatusCardExpired  RTAUStatus = "CardExpiryChanged"
+	RTAUStatusCloseAccount RTAUStatus = "CloseAccount"
+	RTAUExpiryTimeFormat = "1/2006"
```
nirajjayant, line 169: where do we use this?

nirajjayant (approved):

types.go
darshit-bolt, line 169: Let's have this outside of the const block. Ususally when we defined a const block after declaring a type like `RTAUStatus`, we enumerate the constants of that type. The RTAUExpiryTimeFormat is separate.

darshit-bolt (approved):

types.go
namanhkapur: when we get a new expiry date, we need to parse it correctly and update our records; each processor treats expiry dates differently (i.e 1/2006 vs 1-2006 vs 01-06 etc) so creating an RTAU expiry standard that will be our internal representation of expiry updated through RTAU

PR #155, opened 2021-01-04
add rtau processing for adyen gateway

namanhkapur:
- when there's rtau additional data, add the fields into the authorization response
- changing `UpdatedExpiry` to be `time.Time` and making expiry time format processor specific so each processor can parse the time correctly and pass the `time.Time` object (which we'll save in the db if applicable)


types.go
```diff
@@ -105,7 +107,7 @@ type AuthorizationResponse struct {
 	CvvResultRaw         string
 	// Fields in case card details were updated through RTAU
 	RealTimeAccountUpdateStatus RTAUStatus
-	UpdatedExpiry               string
+	UpdatedExpiry               time.Time
 	UpdatedLast4                string
```
dbook13, line 111: Similar to what we do for avs and cvv responses, could we make a new type (`RTAUResponse`) which is a struct containing these fields and then add a field of that type to the AuthorizationResponse? Seems like a more natural way to group these related fields than just using a comment.

dbook13 (changes requested):
Sorry this isn't directly related to the PR. Missed the PR that introduced these fields or I would have left this comment there.

types.go
namanhkapur: oh yeah that makes a lot of sense, way cleaner

gateways/adyen/adyen.go
```diff
@@ -59,7 +60,6 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 	response := &sleet.AuthorizationResponse{
 		TransactionReference: result.PspReference,
 	}
-
 	if result.AdditionalData != nil {
```
dbook13, line 63: Do you think it makes sense to introduce a new function that adds the additional data fields? This if-clause is getting pretty long and making the `Authorize` function pretty long too.

gateways/adyen/adyen.go
```diff
@@ -75,6 +75,23 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 			if cvcRaw, isPresent := values["cvcResultRaw"]; isPresent {
 				response.CvvResultRaw = cvcRaw.(string)
 			}
+			// if rtau status is present, expiry and last 4 must also be present
+			if rtauStatus, isPresent := values["realtimeAccountUpdaterStatus"].(string); isPresent {
+				rtauResponse := sleet.RTAUResponse{
+					RealTimeAccountUpdateStatus: GetRTAUStatus(rtauStatus),
+				}
+				if expiryDate, isPresent := values["expiryDate"].(string); isPresent {
```
dbook13, line 83: I see that expiry and last4 should be present if rtauStatus is present, but, since we check if the expiry and last4 values are present anyway, would the code look cleaner if we had three separate if-clauses for rtauStatus, expiry, and last 4? Or should we return an error if expiry and last4 are not present when rtauStatus is since that goes against our assumption?

gateways/adyen/adyen.go
namanhkapur: yep i'll break it out

gateways/adyen/adyen.go
namanhkapur: yeah that's what i was originally doing but wasn't sure if it was overkill -- I'll have 3 separate if conditions and then a check something like `rtauStatusOk && (!expiryDateOk || !lastFourOk)`

dbook13: What about checking that assumption in hail and logging an error if it's false? That way we don't return an error from sleet and think the transaction failed when in reality it may have been successful but we just got an abnormal rtau response

namanhkapur: okay sounds good -- in hail i can just check the `RTAUResponse` struct

dbook13 (approved):
LGTM, thanks for making all these changes

PR #160, opened 2021-01-27
Authorize.net Apple Pay

dbook13:
Adding network tokenization fields to the Auth.net CreditCard struct and adding logic to populate those fields when an Apple Pay request is sent through sleet.

[Authorize Net Docs](https://developer.authorize.net/api/reference/index.html#payment-transactions-charge-a-tokenized-credit-card)

### DIFF
--- a/common/pointers.go
+++ b/common/pointers.go
@@ -13,3 +13,8 @@ func SafeStr(s *string) string {
 func SPtr(s string) *string {
 	return &s
 }
+
+// BPtr returns a pointer to the bool value b
+func BPtr(b bool) *bool {
+	return &b
+}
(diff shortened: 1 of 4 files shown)


common/pointers.go
```diff
@@ -13,3 +13,8 @@ func SafeStr(s *string) string {
 func SPtr(s string) *string {
 	return &s
 }
+
+// BPtr returns a pointer to the bool value b
+func BPtr(b bool) *bool {
```
SeraYang1, line 18: This functionality already exists in `bolt/pointers/pointers.go`. Not sure why we have two files that do the same thing...

SeraYang1 (approved):

common/pointers.go
SeraYang1: Nvm - this is sleet lol

PR #166, opened 2021-04-26
adding contact card account holder type in sleep

rohanr-bolt:
This is one of the responses that Adyen returns for rtau, and we should close the account in hail if we get this response


dbook13 (approved):
Can you add a description with some context as to why we need this value?

PR #169, opened 2021-05-25
[Adyen Additional Info] Add Recurring Fields for Adyen

ellenwei:
Set field `ShopperReference` on `PaymentRequest` - use `ClientTransactionReference` as the unique identifier. 
Parse the fields for recurring info - `RecurringDetailReference`, `ShopperReference` & `Alias` on response. 
Update unit tests.

### DIFF
--- a/gateways/adyen/adyen.go
+++ b/gateways/adyen/adyen.go
@@ -1,11 +1,12 @@
 package adyen
 
 import (
+	"net/http"
+
 	"github.com/BoltApp/sleet"
 	"github.com/BoltApp/sleet/common"
 	"github.com/adyen/adyen-go-api-library/v4/src/adyen"
 	adyen_common "github.com/adyen/adyen-go-api-library/v4/src/common"
-	"net/http"
 )
 
 // AdyenClient represents the authentication fields needed to make API Requests for a given environment
@@ -159,7 +160,7 @@ func addAdditionalDataFields(
 		response.CvvResultRaw = cvcRaw.(string)
 	}
 
-	// get additional recurring info
+	// set additional recurring info on response
 	if recurringDetailsReference, isPresent := additionalData["recurring.recurringDetailReference"]; isPresent {
 		response.RecurringDetailReference = recurringDetailsReference.(string)
 	}
--- a/gateways/adyen/adyen.go
+++ b/gateways/adyen/adyen.go
@@ -3,6 +3,8 @@ package adyen
 import (
 	"net/http"
 
+	"github.com/google/uuid"
+
 	"github.com/BoltApp/sleet"
 	"github.com/BoltApp/sleet/common"
 	"github.com/adyen/adyen-go-api-library/v4/src/adyen"
@@ -51,7 +53,7 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 	)
 
 	// potentially do something with http response
-	result, _, err := adyenClient.Checkout.Payments(buildAuthRequest(request, client.merchantAccount))
+	result, _, err := adyenClient.Checkout.Payments(buildAuthRequest(request, client.merchantAccount, uuid.New().String()))
 	if err != nil {
 		return &sleet.AuthorizationResponse{
 			Success:              false,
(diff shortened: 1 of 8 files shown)


gateways/adyen/request_builder.go
```diff
@@ -60,6 +60,10 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 		},
 		MerchantAccount:        merchantAccount,
 		MerchantOrderReference: authRequest.MerchantOrderReference,
+
+		// reference to uniquely identify shopper
+		// https://docs.adyen.com/api-explorer/#/CheckoutService/latest/payments__reqParam_shopperReference
+		ShopperReference: *authRequest.ClientTransactionReference,
```
dbook13, line 66: I think ClientTransactionReference will be unique to each transaction rather than to each shopper. Should we use some new reference since we are also using ClientTransactionReference on line 53?

types.go
```diff
@@ -97,15 +97,18 @@ type AuthorizationResponse struct {
...
+	TransactionReference     string
+	AvsResult                AVSResponse
+	CvvResult                CVVResponse
+	Response                 string
+	ErrorCode                string
+	AvsResultRaw             string
+	CvvResultRaw             string
+	RTAUResult               *RTAUResponse
+	RecurringDetailReference string
+	ShopperReference         string
```
dbook13, line 110: Since ShopperReference is going to be the same as the shopper reference we pass into sleet, do we need to return ShopperReference from sleet?

types.go
```diff
@@ -97,15 +97,18 @@ type AuthorizationResponse struct {
...
+	Success                  bool
+	TransactionReference     string
+	AvsResult                AVSResponse
+	CvvResult                CVVResponse
+	Response                 string
+	ErrorCode                string
+	AvsResultRaw             string
+	CvvResultRaw             string
+	RTAUResult               *RTAUResponse
+	RecurringDetailReference string
```
dbook13, line 109: `RecurringDetailReference` and `Alias` are fields that are both very specific to Adyen, so I'm not sure if we should define them at the top level of `AuthorizationResponse`. Would it make more sense to return a metadata map containing these values similar to how we pass an Options map in the `AuthorizationRequest` object?

nirajjayant: Yes I would agree with this, then we can consume that on the hail side. Lets make sure we document it accordingly

types.go
nirajjayant: Yeah maybe we should have the sleet.AuthRequest take in a notion of a ShopperReference/ShopperID and then we don't need to necessarily return it back too

gateways/adyen/request_builder.go
ellenwei: Created UUID as ShooperReference

types.go
ellenwei: It is returned back from adyen along with RecurringDetailReference & Alias in the additionalData map. And since now ShopperReference is using UUID maybe just easier to set from adyen response

types.go
ellenwei: Created map `AdyenAdditionalData`

go.sum
```diff
@@ -106,6 +106,7 @@ github.com/google/pprof v0.0.0-20200229191704-1ebb73c60ed3/go.mod h1:ZgVRPoUq/hf
 github.com/google/pprof v0.0.0-20200430221834-fc25d7d30c6d/go.mod h1:ZgVRPoUq/hfqzAqh7sHMqb3I9Rq5C59dIz2SbBwJ4eM=
 github.com/google/pprof v0.0.0-20200708004538-1a94d8640e99/go.mod h1:ZgVRPoUq/hfqzAqh7sHMqb3I9Rq5C59dIz2SbBwJ4eM=
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
+github.com/google/uuid v1.1.2 h1:EVhdT+1Kseyi1/pUmXKaFxYsDNy9RQYkMWRH68J/W7Y=
```
nirajjayant, line 109: do we need this UUID thing anymore?

gateways/adyen/adyen.go
```diff
@@ -159,6 +160,17 @@ func addAdditionalDataFields(
 		response.CvvResultRaw = cvcRaw.(string)
 	}
 
+	// set additional recurring info on response
+	if recurringDetailsReference, isPresent := additionalData["recurring.recurringDetailReference"]; isPresent {
```
nirajjayant, line 164: can we break this in a similar function like RTAU? We can scope it to just recurring detail reference

nirajjayant (approved):
Is it possible to unit test with the recurring detail reference too? We might need to update the settings on that Adyen merchant.

Won't block for now but something good to have in a follow up?

ellenwei:
> Is it possible to unit test with the recurring detail reference too? We might need to update the settings on that Adyen merchant.
> 
> Won't block for now but something good to have in a follow up?

yea can add unit test and integration test. i will create a separate task

PR #174, opened 2021-06-02
[Citi PLCC] Update for Citi PLCC

ellenwei:
Set RecurringProcessingModel & ShopperInteraction


nirajjayant (commented):
Lets change this, lets have it always addPaymentSpecific fields and then similar to below, we overwrite specifically for CitiPLCC

Does that make sense?

```
addPaymentSpecificFields()

if  authRequest.CreditCardNetwork == CitiPLCC {
   change interaction
}
```

Can we also add a unit test?

nirajjayant (approved):

PR #176, opened 2021-06-21
Add checks for errorInformation in CyberSource Capture/Refund/Void requests

jakechiang-bolt:
There are two sources of errors in CyberSource API responses: Either as a top level field (as `status` and `reason`) or under `errorInformation` (as `errorInformation.Reason` and `errorInformation.Message`). This adds checks for the second case for Capture, Refund,and Void requests.

Also remove duplicate "/" from request path prefix.

Task: https://app.asana.com/0/1148378495214033/1200494589860753/f

### DIFF
--- a/gateways/cybersource/cybersource.go
+++ b/gateways/cybersource/cybersource.go
@@ -100,7 +100,7 @@ func (client *CybersourceClient) Capture(request *sleet.CaptureRequest) (*sleet.
 	if err != nil {
 		return nil, err
 	}
-	capturePath := authPath + "/" + request.TransactionReference + "/captures"
+	capturePath := authPath + request.TransactionReference + "/captures"
 	cybersourceResponse, err := client.sendRequest(capturePath, cybersourceCaptureRequest)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


gateways/cybersource/cybersource.go
```diff
@@ -100,16 +100,22 @@ func (client *CybersourceClient) Capture(request *sleet.CaptureRequest) (*sleet.
 	if err != nil {
 		return nil, err
 	}
-	capturePath := authPath + "/" + request.TransactionReference + "/captures"
+	capturePath := authPath + request.TransactionReference + "/captures"
```
dbook13, line 103: Should we fail for capture/void/refund if `TransactionReference` is empty rather than sending a request that we know is going to fail?

dbook13 (approved):

gateways/cybersource/cybersource.go
ellenwei: This should also fix the double / mentioned here? https://boltpay.atlassian.net/browse/EN-4932

jakechiang-bolt: @dbook13 Yeah that makes sense, what do you think about putting that validation logic in Sleet vs. keeping it outside? My only thought about putting it here is it would need to be repeated for all verbs for each gateway.

@ellenwei Yes, `authPath` ends in a / so this will de-dupe it.

nirajjayant (approved):

PR #179, opened 2021-07-30
[cybs init flags] Add initiator flag information to cybersource payment request

smahbod2014:


gateways/cybersource/request_builders.go
```diff
@@ -6,7 +6,32 @@ import (
...
+	sleet.ProcessingInitiatorTypeStoredMerchantInitiated:   InitiatorTypeMerchant,
+	sleet.ProcessingInitiatorTypeFollowingRecurring:        InitiatorTypeMerchant,
+}
+
 func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error) {
+	initiatorType := initiatorTypeToInitiatorType[*authRequest.ProcessingInitiator]
+	credentialStoredOnFile := false
+	storedCredentialUsed := false
+	if authRequest.ProcessingInitiator != nil {
+		if *authRequest.ProcessingInitiator == sleet.ProcessingInitiatorTypeInitialCardOnFile || *authRequest.ProcessingInitiator == sleet.ProcessingInitiatorTypeInitialRecurring {
```
kenbolt, line 28: can we have a mapping or an array here?

gateways/cybersource/request_builders.go
```diff
@@ -6,7 +6,32 @@ import (
...
+			credentialStoredOnFile = true
+		} else {
```
kenbolt, line 30: same thing, mapping or array rather than else

kenbolt (commented):
lgtm, just use a mapping so that new cases would fail rather than have a default behavior

kenbolt (approved):
thanks!

PR #182, opened 2021-08-20
Add 3DS data to AuthorizationRequest and add to Adyen auths

jakechiang-bolt:
Add field for 3DS verification data to `AuthorizationRequest` and update Adyen request builder to attach this data to auth requests if present.

### DIFF
--- a/gateways/adyen/request_builder.go
+++ b/gateways/adyen/request_builder.go
@@ -100,8 +100,8 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 		request.AdditionalData = buildLevel3Data(level3)
 	}
 
-	// Attach results of 3DS verification was performed
-	if authRequest.ThreeDS != nil {
+	// Attach results of 3DS verification if performed (and not "R"ejected)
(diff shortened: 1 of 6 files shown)


Connor-Bolt:
https://cardinaldocs.atlassian.net/wiki/spaces/CCen/pages/1622573061/Lookup+Response+cmpi+lookup
The PARes Status field has one value we should consider before sending for auth:
`R - Authentication Rejected (Merchant must not submit for authorization)`

gateways/adyen/request_builder.go
```diff
@@ -100,6 +100,23 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 		request.AdditionalData = buildLevel3Data(level3)
 	}
 
+	// Attach results of 3DS verification if performed (and not "R"ejected)
+	if authRequest.ThreeDS != nil && authRequest.ThreeDS.PAResStatus != "R" {
```
andrw, line 104: nit - is there an enum for "R" we can use to compare to?

andrw (approved):

PR #184, opened 2021-09-14
Adding shopperIP, shopperEmail, and ShippingAddress to generic Auth Request + adding support for fields in Ayden

rlcooper46:
Adding ShippingAddress to Auth request
Refactoring BillingAddress type to reflect it being generic Address
Supporting new fields for Ayden request (ShopperIP, ShopperEmail, ShippingAddress)

### DIFF
--- a/gateways/adyen/request_builder.go
+++ b/gateways/adyen/request_builder.go
@@ -66,18 +66,11 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 		ShopperReference: authRequest.ShopperReference,
 	}
 
-	if authRequest.BillingAddress != nil {
-		request.BillingAddress = &checkout.Address{
-			City:              common.SafeStr(authRequest.BillingAddress.Locality),
-			Country:           common.SafeStr(authRequest.BillingAddress.CountryCode),
-			HouseNumberOrName: common.SafeStr(authRequest.BillingAddress.StreetAddress2),
-			PostalCode:        common.SafeStr(authRequest.BillingAddress.PostalCode),
-			StateOrProvince:   common.SafeStr(authRequest.BillingAddress.RegionCode),
-			Street:            common.SafeStr(authRequest.BillingAddress.StreetAddress1),
-		}
-	}
-
+	addBillingAddress(authRequest, request)
 	addPaymentSpecificFields(authRequest, request)
+	addShopperIP(authRequest, request)
+	addShopperEmail(authRequest, request)
+	addShippingAddress(authRequest, request)
 
 	// overwrites the flag transactions
 	if authRequest.ProcessingInitiator != nil {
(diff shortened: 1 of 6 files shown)


rlcooper46:
Still need to add integration tests (waiting for credentials)

gateways/adyen/request_builder.go
```diff
@@ -152,6 +145,48 @@ func addPaymentSpecificFields(authRequest *sleet.AuthorizationRequest, request *
...
+	}
+}
+
+// addShippingAddress adds the shipping address to the Ayden Payment request if available
+func addShippingAddress(authRequest *sleet.AuthorizationRequest, request *checkout.PaymentRequest) {
+	if authRequest.ShippingAddress != nil {
+		request.DeliveryAddress = &checkout.Address{
+			City:              common.SafeStr(authRequest.ShippingAddress.Locality),
+			Country:           common.SafeStr(authRequest.ShippingAddress.CountryCode),
+			HouseNumberOrName: common.SafeStr(authRequest.ShippingAddress.StreetAddress2),
```
rlcooper46, line 182: I copied what we did for billing address but looking at ayden docs maybe we are using these fields incorrectly? Let me know thoughts https://docs.adyen.com/api-explorer/#/CheckoutService/latest/post/payments__reqParam_deliveryAddress

dbook13: Yeah weird it doesn't look like we are using `Street` and `HouseNumberOrName` correctly. @nirajjayant do you know why it was implemented this way?

gateways/adyen/request_builder.go
```diff
@@ -66,18 +66,11 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
...
-			PostalCode:        common.SafeStr(authRequest.BillingAddress.PostalCode),
-			StateOrProvince:   common.SafeStr(authRequest.BillingAddress.RegionCode),
-			Street:            common.SafeStr(authRequest.BillingAddress.StreetAddress1),
-		}
-	}
-
+	addBillingAddress(authRequest, request)
 	addPaymentSpecificFields(authRequest, request)
+	addShopperIP(authRequest, request)
+	addShopperEmail(authRequest, request)
```
dbook13, line 72: Combine into `addShopperData` or `addShopperFields`?

gateways/adyen/request_builder_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build unit
```
dbook13, line 1: How does this differ from the line below? Can we just use one of them?

types.go
```diff
@@ -74,13 +75,16 @@ type Level3Data struct {
...
+	BillingAddress             *Address
 	Level3Data                 *Level3Data
 	ClientTransactionReference *string // Custom transaction reference metadata that will be associated with this request
 	Channel                    string  // for Psps that track the sales channel
 	Cryptogram                 string  // for Network Tokenization methods
 	ECI                        string  // E-Commerce Indicator (can be used for Network Tokenization as well)
 	MerchantOrderReference     string  // Similar to ClientTransactionReference but specifically if we want to store the shopping cart order id
+	ShippingAddress            *Address
+	ShopperIP                  *string
+	ShopperEmail               *string
```
dbook13, line 87: Should we use email from shipping or billing address instead of creating a new field for this?

types.go
```diff
@@ -74,13 +75,16 @@ type Level3Data struct {
...
-	BillingAddress             *BillingAddress
+	BillingAddress             *Address
 	Level3Data                 *Level3Data
 	ClientTransactionReference *string // Custom transaction reference metadata that will be associated with this request
 	Channel                    string  // for Psps that track the sales channel
 	Cryptogram                 string  // for Network Tokenization methods
 	ECI                        string  // E-Commerce Indicator (can be used for Network Tokenization as well)
 	MerchantOrderReference     string  // Similar to ClientTransactionReference but specifically if we want to store the shopping cart order id
+	ShippingAddress            *Address
+	ShopperIP                  *string
```
dbook13, line 86: Since this is a narrowly used piece of data, would it make sense to pass ShopperIP in Options? If it's needed for other processors down the line, we can pull it out into its own field at that point.

If we add fields to this object every time we need to pass one-off pieces of data, the object size is going to get out of hand.

dbook13 (commented):
Can you change the title to something that fits and add a description? Since Sleet is open source, we want the PRs to be extremely explicit in what they are doing and why.

gateways/adyen/request_builder_test.go
rlcooper46: Its the new format but the formatter is adding both for a few versions
https://stackoverflow.com/questions/10646531/golang-conditional-compilation/67937234#67937234

types.go
rlcooper46: I can see billing address, but probably not shipping.

With limited context I have so far maybe we could pull it from billing address. But I think its best to let whoever is using the library choose.

gateways/adyen/request_builder.go
dbook13: Let's just keep it consistent for now and then add a task to the backlog to evaluate changing how we do this.

nirajjayant: Sorry maybe I am not seeing something obvious - is StreetAddress2 not typical for HouseNumberOrName?

types.go
```diff
@@ -18,8 +18,9 @@ type Amount struct {
 	Currency string
 }
 
-// BillingAddress used for AVS checks for auth calls
-type BillingAddress struct {
+// Address generic address to represent billing address, shipping address, etc.
+// used for AVS checks for auth calls
+type Address struct {
```
nirajjayant, line 23: wait this is going to be a breaking change right? If anyone was using BillingAddress this will now fail

nirajjayant (commented):
If you change struct name will cause a build failure for clients I think - we should be backwards compatible

types.go
```diff
@@ -74,13 +75,15 @@ type Level3Data struct {
 type AuthorizationRequest struct {
```
dbook13, line 75: nit: can you alphabetize this object?

types.go
dbook13: Since the ShopperEmail field we are passing is being used as a risk signal / in a velocity rule by Adyen, I think we should pull it from billing address since billing address ties directly into payments. Also, having three different places to pass a shopper email address is going to lead to confusion.

dbook13 (approved):
Couple last comments to address, but overall looks good.

gateways/adyen/request_builder.go
rlcooper46: According to adyen docs ->
123 Street Name
Mountain View, CA 94043

we would fill in as
HouseNumberOrName: 123
Street: Street Name
City: Mountain View
StateOrProvince: CA
PostalCode: 94043

types.go
rlcooper46: Yes will have to refactor uses as well

PR #186, opened 2021-09-27
Adyen address fix

rlcooper46:
Fixes Adyen address formatting to match their API spec:
https://docs.adyen.com/api-explorer/#/CheckoutService/v67/post/payments__reqParam_billingAddress-houseNumberOrName


gateways/adyen/request_builder.go
```diff
@@ -249,3 +252,19 @@ func addIfNonEmpty(value string, key string, data *map[string]string) {
...
+// extractAdyenStreetFormat extracts adyen street format from generic street address
+//                          returns (streetNumber, streetName) format
+//                          If address does not have leading street number, will return ("", street)
+func extractAdyenStreetFormat(streetAddress string) (string, string) {
+	streetLocation := regexp.MustCompile("^\\d+\\s").FindStringIndex(streetAddress)
+	if streetLocation == nil {
+		return "", streetAddress
+	}
+
+	if len(streetAddress) > streetLocation[1] {
```
nirajjayant, line 265: what is this piece doing? Can we add uit tests for it?

rlcooper46: This is just incase we happen to get the address '123 ' we do not throw an indexing error. 

I'll look into adding unit tests

nirajjayant (approved):
lgtm, work with Daniel on his comments

gateways/adyen/request_builder.go
```diff
@@ -249,3 +252,19 @@ func addIfNonEmpty(value string, key string, data *map[string]string) {
 		(*data)[key] = value
 	}
 }
+
+// extractAdyenStreetFormat extracts adyen street format from generic street address
+//                          returns (streetNumber, streetName) format
+//                          If address does not have leading street number, will return ("", street)
+func extractAdyenStreetFormat(streetAddress string) (string, string) {
+	streetLocation := regexp.MustCompile("^\\d+\\s").FindStringIndex(streetAddress)
```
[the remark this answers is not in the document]
dbook13: 1. If we use one of the find submatch methods ([example](https://pkg.go.dev/regexp#example-Regexp.FindStringSubmatch)) on the regex `^(\d+)\s(.*)`, we can directly extract the street number and street name right?
2. Ah right I forgot about that. Let's use back quotes to make this a raw string literal rather than an interpreted string literal to avoid this confusion and get rid of the double back slashes?

PR #191, opened 2021-12-01
authorize_net ->  l2l3 support

rlcooper46:


gateways/authorizenet/request_builders.go
```diff
@@ -109,3 +114,76 @@ func authentication(merchantName string, transactionKey string) MerchantAuthenti
...
+	for i := 0; i < lineItemCount; i++ {
+		lineItem := &LineItem{
+			ItemId: sleet.TruncateString(authRequest.Level3Data.LineItems[i].CommodityCode, 31),
+			Name: sleet.TruncateString(authRequest.Level3Data.LineItems[i].ProductCode, 31),
+			Description: sleet.TruncateString(authRequest.Level3Data.LineItems[i].Description, 255),
+			Quantity: strconv.FormatInt(authRequest.Level3Data.LineItems[i].Quantity, 10),
+			UnitPrice: strconv.FormatInt(authRequest.Level3Data.LineItems[i].UnitPrice.Amount, 10),
+		}
+
+		// This is absolutely terrible why is json -> xml a thing
```
dbook13, line 172: Completely agree but let's make the comment a bit more formal since this is open source haha.

gateways/authorizenet/request_builders.go
```diff
@@ -109,3 +114,76 @@ func authentication(merchantName string, transactionKey string) MerchantAuthenti
...
+
+func buildLineItemsString(authRequest *sleet.AuthorizationRequest) string {
+	// No more than 30 lineitems
+	lineItemCount := len(authRequest.Level3Data.LineItems)
+	if lineItemCount > 30 {
+		lineItemCount = 30
+	}
+
+	lineItems := "{"
+	for i := 0; i < lineItemCount; i++ {
```
dbook13, line 163:
```suggestion
	for i, lineItem := range authRequest.Level3Data.LineItems {
```
If we do this and then break when `i == 30` at the top of the for loop I feel like the logic is a bit more straightforward.

gateways/authorizenet/request_builders.go
```diff
@@ -109,3 +114,76 @@ func authentication(merchantName string, transactionKey string) MerchantAuthenti
...
+			// We must use a string because authnet allows invalid JSON
+			lineItems += "\"lineItem\":" + string(lineItemByte)
+			if i < lineItemCount - 1 {
+				lineItems += ","
+			}
+		}
+	}
+	lineItems += "}"
+
+	if len(lineItems) > 2 {
```
dbook13, line 185: Can we set a bool in the for loop rather than performing this check here? Why we are checking for len > 2 is a bit obscure at first glance.

gateways/authorizenet/request_builders.go
```diff
@@ -109,3 +114,86 @@ func authentication(merchantName string, transactionKey string) MerchantAuthenti
...
+}
+
+// Authorize net converts json to XML before processing the request. This leads to weird scenarios like repeating json
+// fields. LineItems is one of them so we will build it as a raw string
+func buildLineItemsString(authRequest *sleet.AuthorizationRequest) *string {
+	// No more than 30 lineitems
+	lineItemCount := len(authRequest.Level3Data.LineItems)
+	if lineItemCount > 30 {
+		lineItemCount = 30
+	}
```
dbook13, line 165: nit: not needed anymore

dbook13 (approved):

gateways/authorizenet/request_builders.go
rlcooper46: we still need it so we know when to not put a comma

dbook13: You could just replace the last comma with the closing bracket, but I'm fine with either way

PR #195, opened 2021-12-20
Update Sleet Rocketgate SDK version

yuweihsu01:
Update Sleet Rocketgate SDK version to latest


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/golang/protobuf v1.4.3 // indirect
 	github.com/google/go-cmp v0.5.1
 	github.com/jarcoal/httpmock v1.0.5
-	github.com/rocketgate/rocketgate-go-sdk v0.0.0-20211216201057-1639c0c82738 // indirect
+	github.com/rocketgate/rocketgate-go-sdk v0.0.0-20211217152011-c85f16d1e9f3 // indirect
```
dbook13, line 15: Why is this an indirect requirement?

dbook13 (approved):

go.mod
yuweihsu01: ohh, I run `go get github.com/rocketgate/rocketgate-go-sdk` and `indirect` is auto-generated. And I saw some others have this so I also keep this. Should I remove this?

dbook13: I think we directly use this import and it's not indirectly required by a different import so we can probably get rid of the indirect comment. 

https://golangbyexample.com/direct-indirect-dependency-module-go/

PR #199, opened 2022-01-21
Support Apple pay on Cybersource

yuweihsu01:
Support Apple pay on Cybersource

Test:
- [X] Unit Test
- [X] Manual test for Visa, Mastercard, Amex, Discover locally

Asana: https://app.asana.com/0/1200734063164468/1201212075307301/f


gateways/cybersource/request_builders.go
```diff
@@ -92,6 +114,62 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error)
...
+		}
+
+		request.ProcessingInformation.PaymentSolution = "001"	// Apple Pay
+
+		switch authRequest.CreditCard.Network {
+		case sleet.CreditCardNetworkVisa:
+			request.PaymentInformation.TokenizedCard.Type = CardTypeVisa
+			request.ConsumerAuthenticationInformation = &ConsumerAuthenticationInformation{
+				Xid: authRequest.Cryptogram,
+				Cavv: authRequest.Cryptogram,
```
dbook13, line 137: Why do we want the cryptogram in both fields?

gateways/cybersource/request_builders.go
```diff
@@ -92,6 +114,62 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error)
...
+			}
+
+			// For a 40-byte cryptogram, split the cryptogram into two 20-byte binary values (block A and block B).
+			// Send the first 20-byte value (block A) in the cavv field. Send the second 20-byte value (block B) in
+			// the xid field.
+			if len(authRequest.Cryptogram) == AmexCryptogramMaxLength {
+				request.ConsumerAuthenticationInformation = &ConsumerAuthenticationInformation{
+					Cavv: authRequest.Cryptogram[:AmexCryptogramSplitLength],
+					Xid: authRequest.Cryptogram[AmexCryptogramSplitLength:],
```
dbook13, line 159: Can we define a function to determine Cavv and Xid based off the max/split length so we only have to set ConsumerAuthenticationInformation once? I think abstracting the logic will make this section more readable.

gateways/cybersource/request_builders.go
```diff
@@ -92,6 +114,62 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error)
...
+			request.PaymentInformation.TokenizedCard.Type = CardTypeAmex
+			request.ProcessingInformation.CommerceIndicator = CommerceIndicatorAmex
+			request.ConsumerAuthenticationInformation = &ConsumerAuthenticationInformation{
+				Cavv: authRequest.Cryptogram,
+			}
+
+			// For a 40-byte cryptogram, split the cryptogram into two 20-byte binary values (block A and block B).
+			// Send the first 20-byte value (block A) in the cavv field. Send the second 20-byte value (block B) in
+			// the xid field.
+			if len(authRequest.Cryptogram) == AmexCryptogramMaxLength {
```
dbook13, line 156: What happens if the cryptogram is longer than 40 characters? How long is the Apple Pay cryptogram?

gateways/cybersource/request_builders.go
yuweihsu01: From https://developer.cybersource.com/api/developer-guides/dita-payments/CreatingOnlineAuth/CreatingAuthReqPNT.html#id189DF0WI0HT_CreatingAuthReqPNT, it asks me to set both fields for Visa request.

gateways/cybersource/request_builders.go
yuweihsu01: From https://developer.cybersource.com/api/developer-guides/dita-payments/CreatingOnlineAuth/CreatingAuthReqPNT.html#id189DF0WI0HT_CreatingAuthReqPNT, 
`The cryptogram for the American Express card type is either a 20-byte or a 40-byte binary value. For a 20-byte cryptogram, send the cryptogram in the cavv field.`, it seems Cybersource only handle 20 and 40-byte binary value, so should I fail the request if cryptogram is longer than 40 characters? Or should I assume it not happen?

gateways/cybersource/request_builders.go
```diff
@@ -92,6 +114,62 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest) (*Request, error)
...
 	}
+
+	// Apple Pay request
+	if authRequest.Cryptogram != "" {
+		request.PaymentInformation = &PaymentInformation{
+			TokenizedCard: &TokenizedCard{
+				Number: authRequest.CreditCard.Number,
+				ExpirationYear: strconv.Itoa(authRequest.CreditCard.ExpirationYear),
+				ExpirationMonth: fmt.Sprintf("%02d", authRequest.CreditCard.ExpirationMonth),
+				TransactionType: "1",	// in-app transaction
```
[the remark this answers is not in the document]
yuweihsu01: I think "in-app" is something like app related transaction like applepay, googlepay, samsungpay. But because I can make payment on browser, not sure is it good to call it mobile app?

gateways/cybersource/request_builders.go
dbook13: I can't see the docs but does it want Xid to be populated with the transaction ID rather than the cryptogram? The transaction ID is a separate piece of Apple Pay data

gateways/cybersource/request_builders.go
```diff
@@ -11,6 +14,19 @@ const (
...
+	AmexCryptogramSplitLength = 20
+)
+
+const (
+	TransactionTypeInApp = "1"
+)
+
+const (
+	PaymentSolutionApplepay = "001"
+)
```
dbook13, line 28: nit: can you put all of these in the same const clause?

dbook13 (approved):

gateways/cybersource/request_builders.go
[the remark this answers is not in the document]
dbook13: Yeah do they have a generic "ecommerce" transaction type or something similar that we can use instead?

gateways/cybersource/request_builders.go
yuweihsu01: Yeah, from the doc, it mentioned
```For Visa requests:
Set the commerceIndicator field under the processingInformation object to vbv or internet.
Set the xid field under the consumerAuthenticationInformation object to the 3D Secure cryptogram of the payment network token.
Set the cavv field under the consumerAuthenticationInformation object to the 3D Secure cryptogram of the payment network token.
```

And from Amex, it actually also use Xid for Cyrptogram when length is 40-byte. So I think Cybersource would also use Xid for Cryptogram for some use case?

gateways/cybersource/request_builders.go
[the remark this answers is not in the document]
yuweihsu01: So from the doc (https://developer.cybersource.com/api/developer-guides/dita-payments/CreatingOnlineAuth/CreatingAuthReqPNT.html#id189DF0WI0HT_CreatingAuthReqPNT): it just mentioned
`Include the payment network transaction type field, transactionType, under the tokenizedCard object. The only supported value is 1, and it indicates that this is an in-app transaction.`

Also from API doc (https://developer.cybersource.com/api-reference-assets/index.html#payments_payments_process-a-payment_requestexpandedfielddescription_paymentInformation_tokenizedCard):
it explained
```transactionType
Type stringMaxLength 1
Type of transaction that provided the token data. This value does not specify the token service provider; it specifies the entity that
provided you with information about the token.

Possible value:

2: Near-field communication (NFC) transaction. The customer’s mobile device provided the token data for a contactless EMV transaction. For recurring
transactions, use this value if the original transaction was a contactless EMV transaction.
NOTE No CyberSource through VisaNet acquirers support EMV at this time.

Required field for PIN debit credit or PIN debit purchase transactions that use payment network tokens; otherwise, not used.
```
where is not very clear what option else I can use

PR #200, opened 2022-01-25
Checkout.com Integration

cfogg-bolt:
Adds new integration with Checkout.com using their golang SDK.

Notes:
- Includes types+translation for cvv and avs codes but leaves translation disabled. Added comments with the code to add when we want to turn this on.
- Declined to use their public key or idempotent key features, but happy to enable this if we'd find it useful.
- Includes integration tests for Auth, FailedAuth, AuthAndCapture, AuthAndPartialCapture, AuthAndVoid, and AuthCaptureRefund.
- Did not add unit tests since the bulk of the functionality is 1:1 conversions and the checkout.com SDK


integration-tests/checkoutdotcom_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/BoltApp/sleet"
+	"github.com/BoltApp/sleet/gateways/checkoutdotcom"
+	sleet_testing "github.com/BoltApp/sleet/testing"
+	"strings"
+	"testing"
+)
+
+// TestCheckoutDotComAuthorizeFailed
+//
+// Stripe has test cards here: https://stripe.com/docs/testing#cards-responses
```
yuweihsu01, line 13: Checkout.com

integration-tests/checkoutdotcom_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	if !strings.Contains(err.Error(), "Your card has insufficient funds.") {
+		t.Errorf("Response should contain insufficient funds- %s", err.Error())
+	}
+}
+
+// TestCheckoutDotComAuth
+//
+// This should successfully create an authorization on Stripe
```
yuweihsu01, line 32: nit: ditto

arakhshan (approved):

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
+package checkoutdotcom
+
+import (
+	"github.com/BoltApp/sleet"
+	"github.com/checkout/checkout-sdk-go/common"
+	"github.com/checkout/checkout-sdk-go/payments"
+)
+
+func buildChargeParams(authRequest *sleet.AuthorizationRequest) *payments.Request {
```
pranamya-nayak, line 9: good to follow the pattern of returning error.

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+	var source = payments.CardSource{
+		Type: "card",
+		Number: authRequest.CreditCard.Number,
```
pranamya-nayak, line 12: authRequest wouldn't generally have the cc number. All our cc numbers are stored within tokenizer.

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+		ExpiryMonth: uint64(authRequest.CreditCard.ExpirationMonth),
+		ExpiryYear: uint64(authRequest.CreditCard.ExpirationYear),
+		Name: authRequest.CreditCard.FirstName + " " + authRequest.CreditCard.LastName,
```
pranamya-nayak, line 15: did we validate that we have firstname lastname in the CreditCard object. it's usually a part of the billing address.

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+		CVV: authRequest.CreditCard.CVV,
```
pranamya-nayak, line 16: again something present in the tokenizer, we will not get cvv in hail.

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+			City:         *authRequest.BillingAddress.Locality,
+			State:        *authRequest.BillingAddress.RegionCode,
+			ZIP:          *authRequest.BillingAddress.PostalCode,
+			Country:      *authRequest.BillingAddress.CountryCode,
+		},
+	}
+
+	return &payments.Request{
+		Source:   source,
+		Amount:   uint64(authRequest.Amount.Amount),
```
pranamya-nayak, line 29: how about floats?

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+func buildChargeParams(authRequest *sleet.AuthorizationRequest) *payments.Request {
+	var source = payments.CardSource{
+		Type: "card",
+		Number: authRequest.CreditCard.Number,
+		ExpiryMonth: uint64(authRequest.CreditCard.ExpirationMonth),
+		ExpiryYear: uint64(authRequest.CreditCard.ExpirationYear),
+		Name: authRequest.CreditCard.FirstName + " " + authRequest.CreditCard.LastName,
+		CVV: authRequest.CreditCard.CVV,
+		BillingAddress: &common.Address{
+			AddressLine1: *authRequest.BillingAddress.StreetAddress1,
```
pranamya-nayak, line 18: we are already passing in a pointer to authRequest. This would be a pointer to a pointer.

gateways/checkoutdotcom/request_builder.go
```diff
@@ -0,0 +1,59 @@
...
+func buildRefundParams(refundRequest *sleet.RefundRequest) *payments.RefundsRequest {
+	return &payments.RefundsRequest{
+		Amount:    uint64(refundRequest.Amount.Amount),
+		Reference: refundRequest.TransactionReference,
+	}
+}
+
+func buildCaptureParams(captureRequest *sleet.CaptureRequest) *payments.CapturesRequest {
+	return &payments.CapturesRequest{
+		Amount:    uint64(captureRequest.Amount.Amount),
```
pranamya-nayak, line 48: float.

cfogg-bolt: payments.CapturesRequest is from the checkout.com SDK, so they defined the type as unit64

gateways/checkoutdotcom/request_builder.go
rlcooper46: @pranamya-nayak amount should be coming as an integer representing cents so uint64 should be safe

gateways/checkoutdotcom/request_builder.go
rlcooper46: Assuming we follow existing pattern, the CC number will end up being a string that tokenizer recognizes.

This will get substituted into the request by tokenizer after the request gets proxied through it

gateways/checkoutdotcom/request_builder.go
cfogg-bolt: Sorry, still a bit new to golang here. What are you asking me to change?

It seems that the StreetAddress1 is a string pointer and the AddressLine1 is looking for a string (defined by the checkout.com SDK). So I added the * to dereference the string pointer to be a string. When I try removing the *, I get

`Cannot use 'authRequest.BillingAddress.StreetAddress1' (type *string) as the type string`

gateways/checkoutdotcom/request_builder.go
rlcooper46: @pranamya-nayak can confirm this field will be there

pranamya-nayak (approved):
Spoke to Robert to clarify a few things. lgtm

gateways/checkoutdotcom/request_builder.go
cfogg-bolt: Good catch- I double checked the checkout.com docs and cvv isn't a required field, so nil/empty should work. I'll make sure to verify when I get the test credentials. For now I'll still pass here to match other implementations.

gateways/checkoutdotcom/request_builder.go
cfogg-bolt: Added

gateways/checkoutcom/checkoutcom.go
```diff
@@ -0,0 +1,170 @@
...
+
+	input, err := buildCaptureParams(request)
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := checkoutDCClient.Captures(request.TransactionReference, input, nil)
+
+	if err != nil {
+		return &sleet.CaptureResponse{Success: false, ErrorCode: common.SPtr(err.Error())}, nil
```
yuweihsu01, line 95: Curious: should we throw err here if err != nil?

gateways/checkoutcom/request_builder.go
```diff
@@ -0,0 +1,61 @@
...
+			Email: common.SafeStr(authRequest.BillingAddress.Email),
+			Name:  authRequest.CreditCard.FirstName + " " + authRequest.CreditCard.LastName,
+		},
+	}, nil
+}
+
+func buildRefundParams(refundRequest *sleet.RefundRequest) (*payments.RefundsRequest, error) {
+	return &payments.RefundsRequest{
+		Amount:    uint64(refundRequest.Amount.Amount),
+		Reference: *refundRequest.ClientTransactionReference,
```
yuweihsu01, line 44: Do we want to use `TransactionReference` here? (Same for Capture)

gateways/checkoutcom/request_builder.go
```diff
@@ -0,0 +1,61 @@
...
+			Country:      common.SafeStr(authRequest.BillingAddress.CountryCode),
+		},
+	}
+
+	return &payments.Request{
+		Source:   source,
+		Amount:   uint64(authRequest.Amount.Amount),
+		Capture:  common.BPtr(false),
+		Currency: authRequest.Amount.Currency,
+		Reference: *authRequest.ClientTransactionReference,
```
yuweihsu01, line 33: If this is for order reference, then I think we might prefer to use `MerchantOrderReference` (Cart display ID) than `ClientTransactionReference` here? 
cc @dbook13

gateways/checkoutcom/checkoutcom.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Authorize a transaction for specified amount
+func (client *CheckoutComClient) Authorize(request *sleet.AuthorizationRequest) (*sleet.AuthorizationResponse, error) {
+	config, err := checkout.Create(client.apiKey, nil)
+	if err != nil {
+		return nil, err
+	}
+	config.HTTPClient = client.httpClient
+
+	var checkoutDCClient = payments.NewClient(*config)
```
dbook13, line 42: Nit: can we abstract this logic since it's duplicated across all 4 verbs?

dbook13, line 42: Is this a legacy name from before renaming the package to checkoutcom?

gateways/checkoutcom/checkoutcom.go
dbook13: Same question here. I think we want to return err instead of nil here.

gateways/checkoutcom/request_builder.go
dbook13: Agreed. Ideally we would be able to pass both (even if one is in metadata or something), but the merchant will most likely need MerchantOrderReference passed somewhere in order to reconcile orders between their OMS and PSP.

gateways/checkoutcom/request_builder.go
```diff
@@ -0,0 +1,61 @@
...
+func buildRefundParams(refundRequest *sleet.RefundRequest) (*payments.RefundsRequest, error) {
+	return &payments.RefundsRequest{
+		Amount:    uint64(refundRequest.Amount.Amount),
+		Reference: *refundRequest.ClientTransactionReference,
+	}, nil
+}
+
+func buildCaptureParams(captureRequest *sleet.CaptureRequest) (*payments.CapturesRequest, error) {
+	return &payments.CapturesRequest{
+		Amount:    uint64(captureRequest.Amount.Amount),
```
dbook13, line 50: How does checkout.com deal with partial captures and multiple partial captures? Same question for refunds.

dbook13 (approved):
Can you add a link to the checkout.com documentation you used in the description, so devs later on will know where to look if debugging something?

integration-tests/checkoutcom_test.go
```diff
@@ -0,0 +1,187 @@
...
+	if !capture.Success {
+		t.Error("Resulting capture should have been successful")
+	}
+}
+
+// TestCheckoutComAuthPartialCapture
+//
+// This should successfully create an authorization on checkout.com then Capture for a partial amount
+// Since we auth for 1.00USD, we will capture for $0.50
+func TestCheckoutComAuthPartialCapture(t *testing.T) {
```
rlcooper46, line 84: By including this test are we saying we support multicapture with checkout.com?

rlcooper46 (approved):
lgtm, few comments from other's that should be addressed

gateways/checkoutcom/request_builder.go
cfogg-bolt: So transaction reference in this case is a pay id which is included in the url of the request. That's what I'm saving in transaction reference. The reference field they provide is supposed to be for our internal tracking reference, so I used client id here. The checkout.com folks pointed this out when I was passing pay id in the url and the reference field.

cfogg-bolt: Swapped to merchant order ref based on later comments

gateways/checkoutcom/request_builder.go
cfogg-bolt: Yeah they only have one field. Unless we want to do something funky like merging the strings, I'll replace with MerchantOrderReference

gateways/checkoutcom/checkoutcom.go
cfogg-bolt: Yup, updated

gateways/checkoutcom/request_builder.go
cfogg-bolt: There is an extra header you send (capture_type) that indicates whether the capture is final (void any remaining balance) or partial (keep the remaining balance). I don't see this field in the SDK and the default is final.

They're making some other updates to the SDK for us, I'll add in a request to support partial captures too.

integration-tests/checkoutcom_test.go
cfogg-bolt: Yeah I removed it. Won't be able to support until they (or we) updated the SDK to the capture request to add the right headers. Sounds like they'll be updating the SDK soon though.

PR #202, opened 2022-02-01
Add processing initiator parsing to checkout.com integration

cfogg-bolt:
Add processing initiator parsing to checkout.com integration. 

Checkout.com doesn't have a particular field to set based on the processing initiator. Instead, they have a number of flags that are set depending on the processing initiator. The details are in the [docs](https://www.checkout.com/docs/four/payments/accept-payments/use-saved-details/about-stored-card-details), might need a close look to see if I missed something.


gateways/checkoutcom/request_builder.go
```diff
@@ -35,7 +41,35 @@ func buildChargeParams(authRequest *sleet.AuthorizationRequest) (*payments.Reque
 			Email: common.SafeStr(authRequest.BillingAddress.Email),
 			Name:  authRequest.CreditCard.FirstName + " " + authRequest.CreditCard.LastName,
 		},
-	}, nil
+	}
+
+	// see documentation for instructions on stored credentials, merchant-initiated transactions, and subscriptions:
+	// https://www.checkout.com/docs/four/payments/accept-payments/use-saved-details/about-stored-card-details
+	if authRequest.ProcessingInitiator != nil {
```
dbook13, line 48: Would it make sense to move the processor initiator code into its own function?

gateways/checkoutcom/request_builder.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/checkout/checkout-sdk-go/payments"
 )
 
+// Cof specifies the transaction type under the Credential-on-File framework
+const (
+	cofCIT = "CIT"	// Customer Initiated Transaction
+	cofMIT = "MIT"	// Merchant Initiated Transaction
```
dbook13, line 13: It doesn't look like these are used for anything, should we add a const for "Recurring" instead?

dbook13 (approved):

gateways/checkoutcom/request_builder.go
cfogg-bolt: Agreed

gateways/checkoutcom/request_builder.go
cfogg-bolt: Good call, done

PR #207, opened 2022-02-16
Fix cases when amount is less 10 cents for NMI

UnitRoot22:
Slack thread for context: https://boltpay.slack.com/archives/C02STBR02VB/p1645036179039349
When `amountInt` has less than 2 digits i.e is less than 10, the string format logic cannot handle it. I also don't know if this amount will ever be negative, the NMI docs aren't super clear. 
Using this decimal package should fix the issue.

### DIFF
--- a/gateways/nmi/request_builders.go
+++ b/gateways/nmi/request_builders.go
@@ -2,8 +2,10 @@ package nmi
 
 import (
 	"fmt"
-	"github.com/BoltApp/sleet"
 	"strconv"
+
+	"github.com/BoltApp/sleet"
+	"github.com/shopspring/decimal"
 )
 
 // NMI transaction types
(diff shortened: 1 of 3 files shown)


gateways/nmi/request_builders.go
```diff
@@ -2,8 +2,10 @@ package nmi
 
 import (
 	"fmt"
-	"github.com/BoltApp/sleet"
 	"strconv"
+
+	"github.com/BoltApp/sleet"
+	"github.com/shopspring/decimal"
```
UnitRoot22, line 8: There are a few decimal packages with no clear indication of which one is the best. This one as a github projects seems fairly healthy, it also comes with a division function which is ideal.
The other one I checked was https://www.cockroachlabs.com/blog/apd-arbitrary-precision-decimal-package/
I might switch this to use that one instead.

dbook13 (approved):

PR #209, opened 2022-02-22
Check message response codes also to avoid returning success on alrea…

UnitRoot22:
…dy captured messages


gateways/authorizenet/authorizenet.go
```diff
@@ -65,7 +65,8 @@ func (client *AuthorizeNetClient) Capture(request *sleet.CaptureRequest) (*sleet
 		return nil, err
 	}
 
-	if authorizeNetResponse.TransactionResponse.ResponseCode != ResponseCodeApproved {
+	if authorizeNetResponse.TransactionResponse.ResponseCode != ResponseCodeApproved ||
+		isAlreadyCaptured(authorizeNetResponse.TransactionResponse) {
```
UnitRoot22, line 69: If condition becomes any more complex, will split it out into its own set of approved message response codes

gateways/authorizenet/types.go
```diff
@@ -21,6 +21,10 @@ const (
 	ResponseCodeHeld     ResponseCode = "4"
 )
 
+const (
```
UnitRoot22, line 24: We can probably expand this as time goes on, but right now, this is to fix the one issue mentioned in this ticket 
https://app.asana.com/0/1201437212753355/1201868442243928

gateways/authorizenet/test_data/authAlreadyCapturedResponse.json
```diff
@@ -0,0 +1,32 @@
...
+  "messages": {
+    "resultCode": "Ok",
+    "message": [
+      {
+        "code": "I00001",
+        "text": "Successful."
+      }
+    ]
+  }
+}
```
yuweihsu01, line 32: nit: new line

yuweihsu01 (approved):
LGTM

PR #210, opened 2022-02-23
Make sure we return a relevant error code

UnitRoot22:


gateways/authorizenet/authorizenet.go
```diff
@@ -159,6 +159,8 @@ func (client *AuthorizeNetClient) sendRequest(data Request) (*Response, error) {
 func getErrorCode(txnResponse TransactionResponse) string {
 	if len(txnResponse.Errors) > 0 {
 		return txnResponse.Errors[0].ErrorCode
+	} else if len(txnResponse.Messages) > 0 {
```
pranamya-nayak, line 162: don't understand why we would return the message code as error code.

UnitRoot22: So, the message code can still contain error codes. Example, 311 indicates that you are trying to capture a transaction that has already been captured. This is in the message Code. While the transaction response error code will not have anything since the request itself was successful. 
This PR isn't super necessary, but it makes things nicer.

UnitRoot22:
Closing because this is not really necessary.

PR #213, opened 2022-03-03
Added partial capture for checkout.com

cfogg-bolt:
Added partial capture for checkout.com
- updated SDK to pick up partial capture feature
- updated usage of SDK from previous update
- added non-final classification to all checkout.com capture requests. Final voids any remaining authorized amount. Non-final leaves the remaining amount to be voided or captured later.

### DIFF
--- a/gateways/checkoutcom/checkoutcom.go
+++ b/gateways/checkoutcom/checkoutcom.go
@@ -15,26 +15,28 @@ import (
 type CheckoutComClient struct {
 	apiKey     string
 	httpClient *http.Client
+	env        checkout.SupportedEnvironment
 }
 
 const AcceptedStatusCode = 202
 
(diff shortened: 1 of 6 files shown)


rlcooper46 (approved):
lgtm

dbook13 (commented):
What is the level of support that checkout.com has for MPC? Is it supported for all merchants or do merchants have to enable it with checkout.com before we can support it?

cfogg-bolt:
> What is the level of support that checkout.com has for MPC? Is it supported for all merchants or do merchants have to enable it with checkout.com before we can support it?

https://www.checkout.com/docs/four/payments/manage-payments/capture-a-payment doesn't mention merchant settings, but we can ask in the shared channel as well.

PR #215, opened 2022-03-14
Added processing channel id to checkout.com

cfogg-bolt:
Processing Channel ID (PCID) is a new field introduced by checkout.com that it provides to merchants along with an API Key. Merchants can have multiple PCIDs, each mapped to a different payment channel. For more information, see their [docs](https://www.checkout.com/docs/four/resources/api-authentication/api-keys#Multiple_processing_channels).

Since existing merchants only have an API Key and new merchants will be required to provide a processing channel id, the PCID is added as an optional parameter.

Tests were updated to run with and without PCID.


rlcooper46 (approved):

gateways/checkoutcom/request_builder.go
```diff
@@ -38,6 +38,7 @@ func buildChargeParams(authRequest *sleet.AuthorizationRequest) (*payments.Reque
 			Email: common.SafeStr(authRequest.BillingAddress.Email),
 			Name:  authRequest.CreditCard.FirstName + " " + authRequest.CreditCard.LastName,
 		},
+		ProcessingChannelId: common.SafeStr(processingChannelId),
```
rlcooper46, line 41: This will send nil pointer as empty string instead of leaving the field as missing, tests probably confirm if this is fine but something to be aware of

cfogg-bolt: Yeah good catch- in their json conversion, the field is marked as omitempty, so should be ok.

PR #221, opened 2022-03-22
authnet refund add merchant order reference (FIX FROM PREVIOUS REVERTED PR)

rlcooper46:


gateways/authorizenet/types.go
```diff
@@ -105,11 +105,18 @@ type MerchantAuthentication struct {
...
+// ***** AUTHORIZE NET HAS STRICT JSON ORDERING DUE TO XML TRANSLATION *****
+// ***** YOU MUST FOLLOW FIELD ORDER AS SPECIFIED IN THEIR API DOCS    *****
+// ***** https://developer.authorize.net/api/reference/index.html      *****
+// *************************************************************************
+// *************************************************************************
 type TransactionRequest struct {
 	TransactionType TransactionType `json:"transactionType"`
 	Amount          *string         `json:"amount,omitempty"`
 	Payment         *Payment        `json:"payment,omitempty"`
+	RefTransactionID *string         `json:"refTransId,omitempty"`
```
dbook13, line 119: nit: can you fix the indentation?

integration-tests/authorizenet_test.go
```diff
@@ -173,4 +174,47 @@ func TestAuthNetAuthVoid(t *testing.T) {
...
+	}
+
+	refund, err := client.Refund(refundRequest)
+	if err != nil {
+		t.Error("refund request should not have failed")
+	}
+	if !refund.Success {
+		t.Error("Resulting refund should have been successful")
+	}
+}
```
dbook13, line 220: Nit: can you add a new line?

integration-tests/authorizenet_test.go
```diff
@@ -173,4 +174,47 @@ func TestAuthNetAuthVoid(t *testing.T) {
 	}
 }
 
-// TestAuthNetAuthCaptureRefund cannot be written since there is no way to hack settlement and funds cannot refund until they settle
+// TestAuthNetAuthCaptureRefund
+func TestAuthNetAuthCaptureRefund(t *testing.T) {
```
dbook13, line 178: Why perform an auth and capture before the refund if we are just refunding directly against a credit card rather than refunding a specific transaction?

integration-tests/authorizenet_test.go
```diff
@@ -173,4 +174,47 @@ func TestAuthNetAuthVoid(t *testing.T) {
...
+	}
+	if !capture.Success {
+		t.Error("Resulting capture should have been successful")
+	}
+
+	// Refunds for AuthNet take 24 hours to settle. The only option for immediate testing is to do a non-transaction
+	// referenced refund.
+	refundRequest := &sleet.RefundRequest{
+		Amount:               &authRequest.Amount,
+		Last4: authRequest.CreditCard.Number,
```
dbook13, line 206: Are we just overloading last4 with the full number for this test? If so, can we call this out in the comment on line 202?

dbook13 (approved):

integration-tests/authorizenet_test.go
rlcooper46: still need a valid transactionReference if we want to test authnet accepting that field

dbook13: Are we planning to test that later? It doesn't look like we pass it in the refundRequest currently.

rlcooper46: yeah we will once automatic settlement is possible

PR #223, opened 2022-04-14
Treat held in review response as success

lwu-bolt:
Treat `responseCode = 4` as success for auth.net transactions

### DIFF
--- a/gateways/authorizenet/authorizenet.go
+++ b/gateways/authorizenet/authorizenet.go
@@ -47,7 +47,7 @@ func (client *AuthorizeNetClient) Authorize(request *sleet.AuthorizationRequest)
 	}
 
 	return &sleet.AuthorizationResponse{
-		Success:              txnResponse.ResponseCode == ResponseCodeApproved,
+		Success:              txnResponse.ResponseCode == ResponseCodeApproved || txnResponse.ResponseCode == ResponseCodeHeld,
 		TransactionReference: txnResponse.TransID,
 		AvsResult:            translateAvs(txnResponse.AVSResultCode),
 		CvvResult:            translateCvv(txnResponse.CVVResultCode),


dbook13 (commented):
Why do we want to tread "in review" as successful? It seems like "in review" could mean either successful or unsuccessful depending on the result of the review.

lwu-bolt:
This is only for fixing the auth.net incident. We currently declined all "in review" transactions now, if we leave them in the authorized state then the merchant can choose to void or approve/capture the transactions.

dbook13:
I don't know if "authorized" is the correct Bolt state to model an auth.net transaction that is "in review." We will know if a merchant has approved the auth based on if they try to capture, but how will we know if they declined the auth? Additionally, treating this as a successful auth will kick off a capture attempt for merchants with auto-capture enabled. Will that capture fail and how will that impact the transaction flow for a merchant?

lwu-bolt:
Sorry, "authorized state" was not the proper word, what we want to do is to NOT decline the transaction if we received an "in review" response, so the first step is to not mark such transactions as failure (this PR), then seek a way to sync the status with bigC, this is the [task](https://app.asana.com/0/1200734063164468/1202132056317832)

Re: "in review" state across more PSPs, @ellenwei has already started working on auditing other PSPs and see if we need to build a solution for any of them.

lwu-bolt:
The corresponding change of async updating the transaction status has been merged: https://github.com/BoltApp/source/pull/31682

ellenwei:
@dbook13 So the intention here, as @lwu-bolt said, for bigC auth.net, for the txn in review, we failed them immediately, which caused the status not in sync among BigC merchant / bolt / auth.net. 

Currently we are reusing the existing [stripe_radar_post_auth_job.go](https://github.com/BoltApp/source/pull/31682/files#diff-8fb6c8330b82d14cd50a92f5525a4a4c0534494667af362681f4129d518b5e00) to submit review hooks for auth.net to sync status, we also plan to audit all psps to identify similar issues

ellenwei (approved):

lwu-bolt:
Do not decline auth.net responseCode = 4 transactions

pranamya-nayak (approved):
lgtm

PR #226, opened 2022-04-22
Allow MPC support for cybersource

UnitRoot22:
We need to send 2 additional fields if we are doing multiple partial captures for Cybersource.
This is specific to first data north but that logic will live in hail.

### DIFF
--- a/gateways/cybersource/request_builders.go
+++ b/gateways/cybersource/request_builders.go
@@ -179,12 +179,16 @@ func buildCaptureRequest(captureRequest *sleet.CaptureRequest) (*Request, error)
 			Value: *captureRequest.ClientTransactionReference,
 		})
 	}
-	if captureRequest.Options[captureSequenceNumber] != nil && captureRequest.Options[totalCaptureCount] != nil {
-		request.ProcessingInformation = &ProcessingInformation{
-			CaptureOptions: &CaptureOptions{
-				CaptureSequenceNumber: captureRequest.Options[captureSequenceNumber].(string),
-				TotalCaptureCount:     captureRequest.Options[totalCaptureCount].(string),
-			},
+	captureSeqNum, ok := captureRequest.Options[captureSequenceNumber]
+	if ok {
+		totalCapCount, ok := captureRequest.Options[totalCaptureCount]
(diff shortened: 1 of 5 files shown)


gateways/cybersource/request_builders.go
```diff
@@ -169,10 +175,18 @@ func buildCaptureRequest(captureRequest *sleet.CaptureRequest) (*Request, error)
 	}
 	if captureRequest.ClientTransactionReference != nil {
 		request.MerchantDefinedInformation = append(request.MerchantDefinedInformation, MerchantDefinedInformation{
-			Key: "1",
+			Key:   "1",
 			Value: *captureRequest.ClientTransactionReference,
 		})
 	}
+	if captureRequest.Options[captureSequenceNumber] != nil && captureRequest.Options[totalCaptureCount] != nil {
```
rlcooper46, line 182: should we use the val, ok = map[key] pattern instead? A non-pointer interface might not give a nil value

gateways/cybersource/request_builders.go
```diff
@@ -169,10 +175,18 @@ func buildCaptureRequest(captureRequest *sleet.CaptureRequest) (*Request, error)
...
+		request.ProcessingInformation = &ProcessingInformation{
+			CaptureOptions: &CaptureOptions{
+				CaptureSequenceNumber: captureRequest.Options[captureSequenceNumber].(string),
```
rlcooper46, line 185: val,ok = value.(string) and check the value of ok first?

types.go
```diff
@@ -79,17 +79,17 @@ type Level3Data struct {
 type AuthorizationRequest struct {
 	Amount                        Amount
 	BillingAddress                *Address
-	Channel                       string                   // for PSPs that track the sales channel
-	ClientTransactionReference    *string                  // Custom transaction reference metadata that will be associated with this request
+	Channel                       string  // for PSPs that track the sales channel
```
rlcooper46, line 82: not sure why lint messed up our formatting here if we could fix it :)

rlcooper46 (approved):
few comments, buts lgtm after those!

PR #230, opened 2022-05-03
Add pending review for Cybersource

ellenwei:
PENDING_REVIEW: refers to a transaction that has gone through the authorization process and has been flagged and placed in a review queue to be pushed forward for settlement.

Instead of reject the auth request, we will mark it as true, and later in post auth risk review job to send out notification to BigC merchants.

More context: https://app.asana.com/0/0/1202179507225216/f


pranamya-nayak (approved):
lgtm

PR #231, opened 2022-05-23
Add http status code to AuthorizationResponse

lwu-bolt:
Description: this PR adds the http status code and response headers to the `sleet.AuthorizationResponse` struct, they will be passed back to the backend service to identify PSP internal issues.

### DIFF
--- a/gateways/adyen/adyen.go
+++ b/gateways/adyen/adyen.go
@@ -50,12 +50,12 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 	},
 	)
 
-	// potentially do something with http response
 	result, httpResp, err := adyenClient.Checkout.Payments(buildAuthRequest(request, client.merchantAccount))
 	var statusCode int
-	if httpResp != nil && !sleet.IsTokenizerProxyError(httpResp.Header) {
+	if httpResp != nil {
 		statusCode = httpResp.StatusCode
 	}
+	responseHeader := sleet.GetHTTPResponseHeader(request.Options, *httpResp)
 	if err != nil {
 		return &sleet.AuthorizationResponse{
 			Success:              false,
@@ -63,12 +63,14 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 			AvsResult:            sleet.AVSResponseUnknown,
 			CvvResult:            sleet.CVVResponseUnknown,
 			StatusCode:           statusCode,
+			ResponseHeader:       responseHeader,
 		}, err
 	}
 
 	response := &sleet.AuthorizationResponse{
 		TransactionReference: result.PspReference,
 		StatusCode:           statusCode,
+		ResponseHeader:       responseHeader,
 	}
 	if result.AdditionalData != nil {
 		values, ok := result.AdditionalData.(map[string]interface{})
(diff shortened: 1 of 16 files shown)


gateways/adyen/adyen.go
```diff
@@ -51,18 +51,24 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
 	)
 
 	// potentially do something with http response
```
dbook13, line 53: let's remove this comment

utils.go
```diff
@@ -29,3 +36,10 @@ func DefaultIfEmpty(primary string, fallback string) string {
 	}
 	return primary
 }
+
+// IsTokenizerProxyError checks if the error comes from the Tokenizer.
+// It relies on the X-Proxy-Error header which will be set when it's TK internal error:
+// https://github.com/BoltApp/tokenization/blob/master/bolt/tk/app/handlers.ts#L24
+func IsTokenizerProxyError(header http.Header) bool {
+	return len(header.Get(HeaderXProxyError)) > 0
+}
```
dbook13, line 45: This is Bolt-specific logic that probably doesn't belong in sleet since non-Bolt clients of this library will not be proxying through Bolt TK. Can we move it into hail somehow?

dbook13 (changes requested):
Main concern is around the Bolt TK specific logic we are introducing here.

utils.go
lwu-bolt: Added the HTTP response header into the `AuthorizationResponse`, so it can be parsed in hail.

dbook13 (commented):
What do you think about passing back the complete httpResponse to provide the client with additional flexibility? Is the object too heavy to pass back as part of the sleet response?

dbook13:
> What do you think about passing back the complete httpResponse to provide the client with additional flexibility? Is the object too heavy to pass back as part of the sleet response?

Based on some of these processors (like Adyen), it seems like passing back the httpResponse might be pretty standard for these types of API client libraries?

lwu-bolt:
I don't see the necessity to pass back the entire http response:
- the payload could be heavy as you mentioned
- most of the info will be duplicated as the sleet response was parsed from the http response body
- only Adyen returns the raw http response, non of the other API client libraries we use returns it (braintree, checkoutout, rocketgate)

gateways/adyen/adyen.go
```diff
@@ -50,19 +50,27 @@ func (client *AdyenClient) Authorize(request *sleet.AuthorizationRequest) (*slee
...
 	)
 
-	// potentially do something with http response
-	result, _, err := adyenClient.Checkout.Payments(buildAuthRequest(request, client.merchantAccount))
+	result, httpResp, err := adyenClient.Checkout.Payments(buildAuthRequest(request, client.merchantAccount))
+	var statusCode int
+	if httpResp != nil {
+		statusCode = httpResp.StatusCode
+	}
+	responseHeader := sleet.GetHTTPResponseHeader(request.Options, *httpResp)
```
dbook13, line 58: I think we use `request.Options` for other data in Adyen, is there some way to only get the options that are intended to be used to find response headers? Or would it make sense to pass in a function to get desired response headers?

dbook13 (approved):
Approved pending decision on use of `request.Options` for adyen.

lwu-bolt:
Talked offline, it's already extracting a list from the `Options` map, updated ReadMe with usage examples.

PR #239, opened 2022-07-07
Add currency to paypal payflow request

lwu-bolt:
Description: we had an [issue](https://app.asana.com/0/1201437212753355/1202461617728612) where the currency was not correctly shown in paypal, turned out we didn't pass the currency to the request params, this PR is to fix it.

Reference: https://developer.paypal.com/api/nvp-soap/payflow/integration-guide/additional-parameters/#paypal-credit-card-transaction-request-parameters

TestPlan: unit tests + integration tests


rlcooper46 (approved):
lgtm, would suggest adding an integration test that uses another currency (like CAD) to validate if its possible. 

https://github.com/BoltApp/sleet/blob/c37d90254cc29d794d11c0ba17872e2e877e5b36/integration-tests/paypalpayflow_test.go#L119

ihorhorobets-bolt (approved):
LGTM

PR #242, opened 2022-07-20
Add Void On Unsettled Capture Refunds

matthkre:
This allows an option to be sent in the refund request that allows void to be called if the refund fails because its not settled.

### DIFF
--- a/gateways/authorizenet/authorizenet.go
+++ b/gateways/authorizenet/authorizenet.go
@@ -111,6 +111,13 @@ func (client *AuthorizeNetClient) Refund(request *sleet.RefundRequest) (*sleet.R
 	}
 
 	if authorizeNetResponse.TransactionResponse.ResponseCode != ResponseCodeApproved {
+		if shouldVoidUnsettledCapture(request.Options) {
+			for _, errorCode := range authorizeNetResponse.TransactionResponse.Errors {
+				if errorCode.ErrorCode == ErrorCodeIneligibleForIssuingCredit {
+					return sendVoidFromRefundRequest(client, request)
+				}
+			}
+		}
 		errorCode := getErrorCode(authorizeNetResponse.TransactionResponse)
 		response := sleet.RefundResponse{ErrorCode: &errorCode}
 		return &response, nil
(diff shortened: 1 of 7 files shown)


gateways/authorizenet/authorizenet.go
```diff
@@ -121,6 +128,38 @@ func (client *AuthorizeNetClient) Refund(request *sleet.RefundRequest) (*sleet.R
...
+		MerchantOrderReference:     request.MerchantOrderReference,
+	}
+
+	authorzeNetVoidResponse, err := client.Void(&voidRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	if authorzeNetVoidResponse.ErrorCode != nil {
+		return &sleet.RefundResponse{ErrorCode: authorzeNetVoidResponse.ErrorCode}, nil
```
rlcooper46, line 154: Should we also pass Success: False?

gateways/authorizenet/request_builders.go
```diff
@@ -15,7 +15,8 @@ const (
 
 // Options
 const (
-	customerIPOption = "CustomerIP" // Pass as a string pointer
+	customerIPOption                 = "CustomerIP" // Pass as a string pointer
+	shouldVoidUnsettledCaptureOption = "ShouldVoidUnsettledCapture"
```
rlcooper46, line 19: nit, add data type as comment so its clear

rlcooper46 (approved):
couple comments, but overall lgtm

gateways/authorizenet/authorizenet.go
matthkre: This will default to false and seems to be the standard way we structure this.

gateways/authorizenet/request_builders.go
matthkre: Added

gateways/authorizenet/request_builders.go
```diff
@@ -15,7 +15,8 @@ const (
 
 // Options
 const (
-	customerIPOption = "CustomerIP" // Pass as a string pointer
+	customerIPOption                 = "CustomerIP"                 // Pass as a string pointer
+	shouldVoidUnsettledCaptureOption = "ShouldVoidUnsettledCapture" // Pass as a bool
```
dbook13, line 19: I'm a bit worried about sleet clients not understanding that this is potentially a different behavior than the refund they tried to submit if the refund was for less than the total amount of the transaction. At minimum we should add a comment describing how the behaviors different, but I'm still not totally convinced that we should even introduce this into sleet. It seems like processing the response code and determining to submit a void should happen in the client's business logic rather than in sleet just to keep sleet less opinionated.

PR #245, opened 2022-10-19
Fix auth net request nil pointer

lwu-bolt:
Description: fixed an nil pointer panics in auth net request builder

TestPlan: unit tests


jakechiang-bolt (approved):
Thanks, looks good

PR #248, opened 2022-11-17
WIP [Sleet] Pass Google Pay token in Adyen Auth Request

anooj-bolt:
Pass Google Pay token in Adyen Gateway Auth Request.

As documented here: https://docs.adyen.com/payment-methods/google-pay/api-only#before-you-begin.

Ticket: https://app.asana.com/0/1200616417881435/1203366263383020/f

### DIFF
--- a/gateways/adyen/request_builder.go
+++ b/gateways/adyen/request_builder.go
@@ -93,11 +93,16 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 	}
 
 	level3 := authRequest.Level3Data
-	if authRequest.Options[googlePayTokenOption] != nil {
+	if authRequest.Options[googlePayTokenOption] != nil && level3 == nil {
 		googlePayData := map[string]interface{}{
 			"payment.token": authRequest.Options[googlePayTokenOption].(string),
 		}
 		request.AdditionalData = googlePayData
+	} else if level3 != nil && authRequest.Options[googlePayTokenOption] != nil {
+		var builtL3Data map[string]string
+		builtL3Data = buildLevel3Data(level3)
+		builtL3Data["payment.token"] = authRequest.Options[googlePayTokenOption].(string)
+		request.AdditionalData = builtL3Data
 	} else if level3 != nil {
 		request.AdditionalData = buildLevel3Data(level3)
 	}
--- a/gateways/adyen/request_builder.go
+++ b/gateways/adyen/request_builder.go
@@ -20,8 +20,9 @@ const (
 
 // Options
 const (
-	applePayTokenOption = "ApplePayToken"
-	shopperIPOption     = "ShopperIP"
+	applePayTokenOption  = "ApplePayToken"
+	googlePayTokenOption = "GooglePayToken"
+	shopperIPOption      = "ShopperIP"
 )
 
 // Shopper Interactions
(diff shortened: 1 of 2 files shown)


gateways/adyen/request_builder.go
```diff
@@ -92,7 +93,12 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
 	}
 
 	level3 := authRequest.Level3Data
-	if level3 != nil {
+	if authRequest.Options[googlePayTokenOption] != nil {
+		googlePayData := map[string]interface{}{
+			"payment.token": authRequest.Options[googlePayTokenOption].(string),
+		}
+		request.AdditionalData = googlePayData
+	} else if level3 != nil {
```
rlcooper46, line 101: I think with this logic, we will skip building the level3 data if google pay is present. Lets adjust to do both

anooj-bolt: We don't want to build the l3 data for Google Pay

rlcooper46: we should send both, its up to adyen to decide if the l2l3 data applies to the transaction

anooj-bolt: Addressed

yuweihsu01 (commented):
I think for Adyen Gateway, it used Checkout API and you should follow https://docs.adyen.com/payment-methods/google-pay/api-only#before-you-begin.

yuweihsu01:
And it seems we are using v65 [[DD](https://app.datadoghq.com/logs?query=%22live.adyenpayments.com%2Fcheckout%2F%22&cols=service%2C%40request_path&event=AQAAAYSHdIHY2TdP_gAAAABBWVNIZEpHMUFBQ0xmU1pzYzZEM3l3QUU&index=main&messageDisplay=inline&stream_sort=time%2Cdesc&viz=stream&from_ts=1668630682826&to_ts=1668717082826&live=true)], but the requirement seems to be v67. Have asked them in email.

gateways/adyen/request_builder.go
```diff
@@ -92,7 +93,17 @@ func buildAuthRequest(authRequest *sleet.AuthorizationRequest, merchantAccount s
...
+		googlePayData := map[string]interface{}{
+			"payment.token": authRequest.Options[googlePayTokenOption].(string),
+		}
+		request.AdditionalData = googlePayData
+	} else if level3 != nil && authRequest.Options[googlePayTokenOption] != nil {
+		var builtL3Data map[string]string
+		builtL3Data = buildLevel3Data(level3)
+		builtL3Data["payment.token"] = authRequest.Options[googlePayTokenOption].(string)
+		request.AdditionalData = builtL3Data
+	} else if level3 != nil {
```
rlcooper46, line 106: nit: we could probably simplify this a bit?
```suggestion
	if authRequest.Options[googlePayTokenOption] != nil {
		googlePayData := map[string]interface{}{
			"payment.token": authRequest.Options[googlePayTokenOption].(string),
		}
		request.AdditionalData = googlePayData
	} 
	if level3 != nil  {
	        if request.AdditionalData == nil {
	            request.AdditionalData = make(map[string]string)
	        }
		buildLevel3Data(level3, request.AdditionalData)
	} 
```

anooj-bolt: actually getting rid of this as its the wrong approach

yuweihsu01 (approved):
LGTM

PR #250, opened 2022-11-29
Add Auth Net Auth Endpoint Support for Google Pay Tokens

anooj-bolt:
Google Pay tokens can now be passed through the sleet request and will build a Google Pay specific request to hit the auth endpoint with

### DIFF
--- a/gateways/authorizenet/request_builders.go
+++ b/gateways/authorizenet/request_builders.go
@@ -1,6 +1,7 @@
 package authorizenet
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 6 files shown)


gateways/authorizenet/request_builders.go
```diff
@@ -37,7 +38,25 @@ func buildAuthRequest(merchantName string, transactionKey string, authRequest *s
...
-		TransactionRequest: TransactionRequest{
+	}
+
+	var transactionRequest TransactionRequest
+	if authRequest.Options[sleet.GooglePayTokenOption] != nil {
+		// Google Pay request
+		googlePayToken := authRequest.Options[sleet.GooglePayTokenOption].(string)
+		encodedGooglePayToken := base64.StdEncoding.EncodeToString([]byte(googlePayToken))
+		transactionRequest = TransactionRequest{
+			TransactionType: TransactionTypeAuthCapture,
```
yuweihsu01, line 49: Just curious, does this mean auto capture? or this is just specific for GooglePay?

yuweihsu01 (approved):
LGTM with one small question

gateways/authorizenet/types.go
```diff
@@ -152,6 +157,15 @@ type Customer struct {
 // Payment specifies the credit card to be authorized (only payment option for now)
 type Payment struct {
 	CreditCard CreditCard `json:"creditCard"`
+	OpaqueData OpaqueData `json:"opaqueData"`
```
rlcooper46, line 160: json order matters for authnet, lets just confirm that this is the correct order.

gateways/authorizenet/request_builders.go
anooj-bolt: Specific to Google Pay. https://developer.authorize.net/api/reference/index.html#mobile-in-app-transactions-create-a-google-pay-transaction.

gateways/authorizenet/types.go
anooj-bolt: OpaqueData shouldn't be there when CreditCard is there

gateways/authorizenet/request_builders.go
yuweihsu01: But based on https://developer.authorize.net/api/reference/features/payment_transactions.html, I think this seems to be Auth + Capture. Would https://developer.authorize.net/api/reference/index.html#mo just list one example? Could you check on this?

yuweihsu01 (approved):

PR #251, opened 2022-11-30
Remove Base64 encode for authnet Gpay token

anooj-bolt:
Authorize Net has wrong documentation.

https://developer.authorize.net/api/reference/features/in-app.html

The token we get back from Google Pay is already base64 encoded.

### DIFF
--- a/gateways/authorizenet/request_builders.go
+++ b/gateways/authorizenet/request_builders.go
@@ -48,7 +48,7 @@ func buildAuthRequest(merchantName string, transactionKey string, authRequest *s
 			Payment: &Payment{
 				OpaqueData: &OpaqueData{
 					DataDescriptor: GooglePayPaymentDescriptor,
-					DataValue:      authRequest.Options[sleet.GooglePayTokenOption].([]byte),
+					DataValue:      authRequest.Options[sleet.GooglePayTokenOption].(string),
 				},
 			},
 		}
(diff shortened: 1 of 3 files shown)


gateways/authorizenet/request_builders.go
```diff
@@ -43,15 +42,13 @@ func buildAuthRequest(merchantName string, transactionKey string, authRequest *s
...
-		googlePayToken := authRequest.Options[sleet.GooglePayTokenOption].(string)
-		encodedGooglePayToken := base64.StdEncoding.EncodeToString([]byte(googlePayToken))
 		transactionRequest = TransactionRequest{
 			TransactionType: TransactionTypeAuthOnly,
 			Amount:          &amountStr,
 			Payment: &Payment{
 				OpaqueData: &OpaqueData{
 					DataDescriptor: GooglePayPaymentDescriptor,
-					DataValue:      encodedGooglePayToken,
+					DataValue:      authRequest.Options[sleet.GooglePayTokenOption].(string),
```
yuweihsu01, line 51: I think you might also need to update your unit test?

yuweihsu01 (approved):

gateways/authorizenet/request_builders.go
anooj-bolt: Forgot to push commit oops

PR #253, opened 2023-02-08
Add auth message and result type for Adyen

jordan-mcgregor:
Support modeled adyen errors in order to better model auths attempts that return these error types - https://docs.adyen.com/development-resources/error-codes#checkout-error-codes

### DIFF
--- a/types.go
+++ b/types.go
@@ -134,6 +134,8 @@ type AuthorizationResponse struct {
 	CvvResult             CVVResponse
 	Response              string
 	ErrorCode             string
+	Message               string // message from the gateway describing the reason for a failed auth
+	ResultType            ResultType
 	AvsResultRaw          string
 	CvvResultRaw          string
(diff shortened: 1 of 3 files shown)


types.go
```diff
@@ -241,3 +243,13 @@ type ThreeDS struct {
 	Version          string // 3DS Version
 	XID              string // Transaction ID from authentication processing (for 3DS1)
 }
+
+type ResultType string
+
+const (
+	ResultTypeSuccess      ResultType = "Approved"
```
yuweihsu01, line 250: curious: so we are not going to use it for now?

yuweihsu01 (approved):

types.go
jordan-mcgregor: Yes its a bit redundant with the `Success` field but would be much bigger effort to completely remodel the auth response and make consistent across the various PSPs.

PR #254, opened 2023-02-21
Fix adyen APIError failure response

jordan-mcgregor:
Don't return an go error when response is correctly modelled.

### DIFF
--- a/gateways/adyen/adyen.go
+++ b/gateways/adyen/adyen.go
@@ -81,7 +81,7 @@ func (client *AdyenClient) AuthorizeWithContext(ctx context.Context, request *sl
 				ErrorCode:  adyenError.Code,
 				Message:    adyenError.Message,
 				ResultType: sleet.ResultTypeAPIError,
-			}, err
+			}, nil
 		}
 		return &sleet.AuthorizationResponse{
 			Success:              false,
--- a/integration-tests/adyen_test.go
+++ b/integration-tests/adyen_test.go
@@ -1,7 +1,6 @@
 package test
 
 import (
-	"strings"
 	"testing"
 
 	"github.com/BoltApp/sleet"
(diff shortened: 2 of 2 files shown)


jordan-mcgregor:
Current dependencies on/for this PR:
* master
  * **PR #254** <a href="https://app.graphite.dev/github/pr/BoltApp/sleet/254" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/BoltApp/sleet/254?utm_source=stack-comment).

gateways/adyen/adyen.go
```diff
@@ -81,7 +81,7 @@ func (client *AdyenClient) AuthorizeWithContext(ctx context.Context, request *sl
 				ErrorCode:  adyenError.Code,
 				Message:    adyenError.Message,
 				ResultType: sleet.ResultTypeAPIError,
-			}, err
+			}, nil
```
yuweihsu01, line 84: Few questions:
1. how about line 94?
2. If we don't return error here, will hail/bolt/payments/adyen/gateway.go always go with `ResultTypePaymentError`? (maybe `ResultTypeServerError` also if line 94 return error )

jordan-mcgregor: 1. Line 94 is a good point, however would make things backwards incompatible for the library. In idiomatic Go, if an error is not nil then the other parameters should be ignored. However this is not the contract that sleet has been observing. All of other the integrations I saw follow this pattern where the error is returned along with some information populated in the auth response. For consistency with other PSPs for now I think best to keep as is. These APIErrors we can model better than an http error though.

2. I have a change for hail/bolt/payments/adyen/gateway.go also but waiting till this PR is merged so I can update to the correct sleet version.

yuweihsu01 (approved):

PR #257, opened 2023-04-03
Update Authorize Net to support Google Pay Refunds

anooj-bolt:
Update Authorize Net to support Google Pay Refunds. Authorize Net requires last 4 digits of credit card number for all refunds. We don't currently store this for auth.net Google Pay transactions, however we have the ability to fetch this from auth net via api(getTransactionDetails). This PR fetches the last4 from getTransactionDetails and populates the CC number in the refund request

Related bug: https://app.asana.com/0/1200616417881435/1204243318709223/f

Thread: https://boltpay.slack.com/archives/C03U3T92Y8Z/p1680433093558869

### DIFF
--- a/gateways/authorizenet/authorizenet.go
+++ b/gateways/authorizenet/authorizenet.go
@@ -4,11 +4,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"io/ioutil"
-	"net/http"
-
 	"github.com/BoltApp/sleet"
 	"github.com/BoltApp/sleet/common"
+	"io/ioutil"
+	"net/http"
 )
 
 var (
@@ -134,6 +133,20 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
 		return nil, err
 	}
 
+	var transactionDetailsResponse *sleet.TransactionDetailsResponse
+	if request.Options != nil && request.Options[sleet.GooglePayTokenOption].(bool) {
+		transactionDetailsResponse, err = client.GetTransactionDetails(&sleet.TransactionDetailsRequest{
+			TransactionReference: request.TransactionReference,
+		})
+		if err != nil {
+			return nil, err
+		}
+		//Past last 4 digits of card number
+		creditCardNumber := transactionDetailsResponse.CardNumber
+		last4 := creditCardNumber[len(creditCardNumber)-4:]
+		authorizeNetRefundRequest.CreateTransactionRequest.TransactionRequest.Payment.CreditCard.CardNumber = last4
+	}
+
 	authorizeNetResponse, _, err := client.sendRequest(ctx, *authorizeNetRefundRequest)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


pranamya-nayak (approved):
lgtm

gateways/authorizenet/authorizenet.go
```diff
@@ -134,6 +135,24 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
 		return nil, err
 	}
 
+	var transactionDetailsResponse *sleet.TransactionDetailsResponse
+	if request.Options != nil && request.Options[sleet.GooglePayTokenOption] != nil {
+		var boolValue bool
+		boolValue, _ = strconv.ParseBool(request.Options[sleet.GooglePayTokenOption].(string))
```
yuweihsu01, line 141: so we do not plan to handle error case?

gateways/authorizenet/authorizenet.go
```diff
@@ -134,6 +135,24 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
...
+				TransactionReference: request.TransactionReference,
+			})
+			if err != nil {
+				return nil, err
+			}
+			//Past last 4 digits of card number
+			creditCardNumber := transactionDetailsResponse.CardNumber
+			last4 := creditCardNumber[len(creditCardNumber)-4:]
+			authorizeNetRefundRequest.CreateTransactionRequest.TransactionRequest.Payment.CreditCard.CardNumber = last4
+		}
```
yuweihsu01, line 153: Any nil pointer needs to be checked here?

gateways/authorizenet/authorizenet.go
```diff
@@ -134,6 +135,24 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
 		return nil, err
 	}
 
+	var transactionDetailsResponse *sleet.TransactionDetailsResponse
+	if request.Options != nil && request.Options[sleet.GooglePayTokenOption] != nil {
+		var boolValue bool
+		boolValue, _ = strconv.ParseBool(request.Options[sleet.GooglePayTokenOption].(string))
+		if boolValue {
+			transactionDetailsResponse, err = client.GetTransactionDetails(&sleet.TransactionDetailsRequest{
+				TransactionReference: request.TransactionReference,
```
yuweihsu01, line 144: Curious: so this would be original transaction?

gateways/authorizenet/authorizenet.go
```diff
@@ -134,6 +135,24 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
 		return nil, err
 	}
 
+	var transactionDetailsResponse *sleet.TransactionDetailsResponse
+	if request.Options != nil && request.Options[sleet.GooglePayTokenOption] != nil {
```
yuweihsu01, line 139: add some unit tests for this?

gateways/authorizenet/request_builders.go
```diff
@@ -260,3 +260,16 @@ func buildLineItemsString(authRequest *sleet.AuthorizationRequest) *string {
 	}
 	return nil
 }
+
+func BuildTransactionDetailsRequest(merchantName string, transactionKey string, transactionDetailsRequest *sleet.TransactionDetailsRequest) (
```
yuweihsu01, line 264: add unit test?

gateways/authorizenet/authorizenet.go
```diff
@@ -134,6 +135,24 @@ func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request
 		return nil, err
 	}
 
+	var transactionDetailsResponse *sleet.TransactionDetailsResponse
```
yuweihsu01, line 138: I think it could be declare inside the `if`

gateways/authorizenet/authorizenet.go
yuweihsu01, line 139: wonder if it would be better to move these logic inside the `buildRefundRequest`? we can query `GetTransactionDetails` before calling `buildRefundRequest` if needed.

gateways/authorizenet/authorizenet.go
anooj-bolt: No this is already guaranteed to be built based in the buildRefundRequest function

gateways/authorizenet/authorizenet.go
anooj-bolt: Yes

gateways/authorizenet/authorizenet.go
anooj-bolt: No I think we want to separate this from buildRefundRequest

yuweihsu01 (approved):

gateways/authorizenet/authorizenet.go
```diff
@@ -129,6 +129,16 @@ func (client *AuthorizeNetClient) Refund(request *sleet.RefundRequest) (*sleet.R
 
 // RefundWithContext refunds a captured transaction with amount and captured transaction reference
 func (client *AuthorizeNetClient) RefundWithContext(ctx context.Context, request *sleet.RefundRequest) (*sleet.RefundResponse, error) {
+	transactionDetailsResponse, err := client.GetTransactionDetails(&sleet.TransactionDetailsRequest{
```
yuweihsu01, line 132: Curious: so is this only used for Google Pay? Maybe we could skip this call for non-GooglePay request?

PR #261, opened 2023-08-25
Fix response fields used by CyberSource tokenization

Wolfizen:
## Description

The documentation I followed [here](https://developer.cybersource.com/docs/cybs/en-us/tms/developer/all/rest/tms-developer/Authorizations-for-Tokens/Authorize-a-Payment-and-Create-Customer-Tokens.html) used an incorrect schema. The real responses observed do not match the documentation.

This PR updates the fields we use to match the real observed responses.

### DIFF
--- a/gateways/cybersource/cybersource.go
+++ b/gateways/cybersource/cybersource.go
@@ -205,10 +205,6 @@ func (client *CybersourceClient) VoidWithContext(ctx context.Context, request *s
 	}
 	voidPath := authPath + request.TransactionReference + "/voids"
 	cybersourceResponse, _, err := client.sendRequest(ctx, voidPath, cybersourceVoidRequest)
-	if cybersourceResponse != nil {
-		b, _ := json.Marshal(cybersourceResponse)
-		fmt.Printf("resp %s\n", b) // debug
-	}
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


Wolfizen:
Current dependencies on/for this PR:
* master
  * **PR #261** <a href="https://app.graphite.dev/github/pr/BoltApp/sleet/261" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/BoltApp/sleet/261?utm_source=stack-comment).

darshit-bolt (approved):

Wolfizen:
I'm completely unsure why CYBS TestVoid is failing.. it works when I run the integration test on my local machine against our sandbox account

Wolfizen:
1. Add debug test
```go
if cybersourceResponse != nil {
		b, _ := json.Marshal(cybersourceResponse)
		fmt.Printf("resp %s\n", b) // debug
	}
```
2. Build passes
3. Remove debug test
4. Build fails
🤯

Wolfizen:
Ok, so,, I added all the logging I could, and it seems that this may have just been a Cybersource sandbox service issue. Nothing looked different between the successes and failures.

I'm going to keep the logging because I put a bunch of work into it and it can be re-used for other tests that want to log their http requests. (The logs aren't part of the non-test code)

PR #265, opened 2024-01-30
[Authorize.net] Various corrections to L2/L3 data inputs

Wolfizen:
Our L2/L3 Line Item inputs were not fully compliant with Authorize.net's specification.

Firstly, we were sending the incorrect item fields. CommodityCode was used for the item ID when instead ProductCode should be used. And ProductCode was used for the name when Description should be used.

Next, Authorize.net requires an item ID and a name on every item. Some cart items may not have a product code or description added. To facilitate successful auths when limited data is available, we will use the item index as a substitute for item ID, and the item ID as substitute for the item name. There is precedent for adding this logic for other L2/L3-supported processors.

Lastly, I renamed some types and fields and added notes regarding the specific requirements for each field from Authorize.net's documentation.

### DIFF
--- a/gateways/authorizenet/request_builders.go
+++ b/gateways/authorizenet/request_builders.go
@@ -184,15 +184,15 @@ func addL2L3Data(
 			authNetAuthRequest.TransactionRequest.LineItem = json.RawMessage(*lineItemString)
 		}
 
-		authNetAuthRequest.TransactionRequest.Tax = &Tax{
+		authNetAuthRequest.TransactionRequest.Tax = &ExtendedAmount{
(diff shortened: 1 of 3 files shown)


Wolfizen:
* `master`
  * **#265** <a href="https://app.graphite.dev/github/pr/BoltApp/sleet/265?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @Wolfizen and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

libo-gao (approved):

PR #269, opened 2024-02-09
[CKO Balance Transfer] - Adding a balance transfer endpoint

ashish-bolt:
Adding a balance transfer endpoint for CKO. With the setup of the interfaces, we will need to incorporate some not implemented methods in the other clients.


darshit-bolt (commented):
I don't think we should add BalanceTransfer method to the Client interface such that it required by all PSPs to implement this. 
A balance transfer is an exclusively payfac concept and by design the sleet client is meant to be very generic for all PSPs. So I think that we are leaking payfac concepts to this client which doesn't look ideal.

I think the checkoutcom gateway should still have the BalanceTransfer public method, and this will be called directly by hail code without using the generic sleet client

